PR #1, opened 2015-03-15
Update readme to point to github

rocketman768:
Also split the website section in twain for devs and users.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -38,13 +38,19 @@ Author list created with:
 
 ## Websites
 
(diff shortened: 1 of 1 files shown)


rocketman768:
@mikfire need some :eyes: for a :+1: or a :-1: cuz you might :heart: it or it might be :hankey:

mikfire:
Looks good. I was somewhat hoping we would use github's problem tracking. But. Umm. No habla emojii

PR #12, opened 2015-04-04
Update build instructions in readme

rocketman768:


malavv:
This is great, the old dependencies were getting old.
:+1:

Issue #25, opened 2015-04-18
Multiple dialogs when cancelling an action for more than one Item in the tree


rocketman768:
From launchpad [#1445671](https://bugs.launchpad.net/brewtarget/+bug/1445671)

When selecting more than one Item in the tree (i.e. recipes) and choose
'copy' from the context-menue, a dialog appears which asks for a name
for each copy. If you decide to abort copying you have to cancel this
dialog for every previously selectet item. Maybe it would be nice if
hitting "cancel" once aborts all of the following copy-processes.

Brewtarget 2.1.0 on MacOS 10.10.3

mikfire:
How do I assign this ticket to myself?

rocketman768:
Directly to the right of the original comment, there is a column with "Labels," "Milestone," and "Assignee" sections. Click on assignee and you should see your id.

mikfire:
The "click on assignee" doesn't work. As in, that isn't something I can click.

rocketman768:
Hm. Let me see if there is some setting I have to change.

mikfire:
Oh. I like that.

PR #26, opened 2015-04-22
Closes feature request #25 -- multiple dialogs

mikfire:
This closes feature request #25 -- multiple dialogs when canceling a copy. I
rearranged things so that the code in MainWindow is much cleaner; there are no
more copyThis() methods. Instead I drove all the complexity into BtTreeView
and BtTreeModel, like deleteSelected.

This does somewhat change the syntax of the cancel button. To avoid copying a
single item, just enter an empty name. I added a tool type to explain this. I
also changed the "Cancel" button to read "Cancel All", hoping this will make
it more obvious.


rocketman768:
:+1:

PR #27, opened 2015-04-26
fixes #2 - Color range wrongly updated after style change

### LINKED ISSUE #2: Color range wrongly updated after style change
As reported by @theophae

I noticed that if have set EBC as default unit in parameters, if you change the style of the beer in the recipe, the color range of the style is not updated correctly. To get the right display, you need to right click on "Color" and change the display unit to another one and to change it again to EBC. After doing this, the color range changes to fit the right value.

theophae:
The problem appears only if you configure EBC as default color unit in configuration because after changing the style, the color range was updated directly in SRM without taking care of the default unit.


rocketman768:
:+1:

PR #33, opened 2015-05-05
Fixes #30 - Recipe doesn't update automatically after yeast properties changes.

### LINKED ISSUE #30: Recipe doesn't update automatically after yeast properties changes.
If you change, for example, the attenuation percentage of the yeast you have already added (different to adding a new yeast in the list), then you click on "Ok", the Fg is not updated.

I have made the changes, you can assign this one to me.

By the way, tell me if I can send my pull request directly without opening an issue. As I don't know yet, I'm doing this way.

theophae:
This fix the bug #30.
The fix is very simple. I just duplicate the mechanism used for the other ingredients.
Please review

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -271,8 +271,9 @@ bool Database::load()
    QHash<int,Recipe*>::iterator i;
    QList<Fermentable*>::iterator j;
    QList<Hop*>::iterator k;
-   QList<Mash*>::iterator l;
-   QList<MashStep*>::iterator m;
+   QList<Yeast*>::iterator l;
+   QList<Mash*>::iterator m;
+   QList<MashStep*>::iterator n;
 
 
    for( i = allRecipes.begin(); i != allRecipes.end(); i++ )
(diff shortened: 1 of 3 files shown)


src/database.cpp
```diff
@@ -290,18 +291,22 @@ bool Database::load()
          connect( *j, SIGNAL(changed(QMetaProperty,QVariant)), *i, SLOT(acceptFermChange(QMetaProperty,QVariant)) );
       
       QList<Hop*> tmpH = hops(*i);
-      for( k = tmpH.begin(); k != tmpH.end(); ++k )
+      for( k = tmpH.begin(); k != tmpH.end(); k++ )
```
rocketman768: Leave as a pre-increment

theophae:
Ok, I'll change it. Thanks for the explanation :-)

theophae:
Ok, I restored the pre-increment, and I change the post-increment use for fermentable to a pre-increment for the same reason.
Thanks for the explanation :-)

rocketman768:
:+1:

Great! Squash the 2 commits into 1 and I will merge (see [here](http://davidwalsh.name/squash-commits-git)).

theophae:
Squash successfully done (after several fails ^^). You can merge it know

PR #35, opened 2015-05-05
Fixes #29 - Bug in scale recipe calculation by efficiency

### LINKED ISSUE #29: Bug in scale recipe calculation by efficiency
I found a major issue in the scale recipe calculation tool. If you use some sugar or some extract in you recipe and you want to change the recipe efficiency by using the scale recipe tool, the original gravity will change whereas it should not. This is not noticeable if you don't have a lot of non mashable sugar in you recipe or if you change the efficiency for a few point. But this bug is very noticeable if you go from 70% efficiency to 50% for instance.

Enough talking ! I found this bug, but I also found the fix ;-) so please assign it to me.

theophae:
This fixes the bug #29 .
Please review

### DIFF
--- a/src/ScaleRecipeTool.cpp
+++ b/src/ScaleRecipeTool.cpp
@@ -110,7 +110,10 @@ void ScaleRecipeTool::scaleByEfficiency()
       if( ferm == 0 )
          continue;
 
-      ferm->setAmount_kg(ferm->amount_kg() * ratio);
+      if( !ferm->isSugar() && !ferm->isExtract() )
+      {
+         ferm->setAmount_kg(ferm->amount_kg() * ratio);
+      }
    }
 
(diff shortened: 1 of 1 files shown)


theophae:
Maybe my fix is not complete. I wonder about the case of steeped grain, should it be added to this condition ? Because, to me, the efficiency is only about mashing process so not-mashed grain amount should not change.
What do you think ?

rocketman768:
The efficiency parameter we use is overall efficiency, so I think this is likely a sufficient fix.

We used to use mash efficiency in early brewtarget versions, and use the volume losses and other parameters to estimate the overall efficiency, but the calculations get very complicated and sometimes recursive. I think that's why other software also use the overall efficiency :smiley:

rocketman768:
@Brewtarget/developers have a look

Issue #39, opened 2015-05-14
Can not export recipe


Mackanzx:
When exporting recipes, converting the filename QString to ISO-8859-1 seems to break the string, in my environment at least.
"toAscii" of QString was deprecated in QT5, and it is recommended to be replaced with "toLatin1", which was done in commit 6f6315dd086c89603645d82b37ec428dbb9213f4. In my environment, toLatin1() produces this string:
    ´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢┐¢´┐¢´┐¢´┐¢´┐¢´┐¢´┐¢i      s0@´┐¢T∟´┐¢´┐¢

I propose to not convert the QString to anything, since setFileName of QFile takes a QString.

Pull request: https://github.com/Brewtarget/brewtarget/pull/40

PR #40, opened 2015-05-14
issue:#39 - Can not export recipe.

Mackanzx:
Fixing issue https://github.com/Brewtarget/brewtarget/issues/39 : Not unnecessarily converting the QString filename before feeding it to a QFile.


rocketman768:
:+1: nice catch

PR #42, opened 2015-05-20
Add post-boil loss OG test

rocketman768:
Make a test verifying post-boil losses do not affect the OG

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -422,7 +422,10 @@ ADD_TEST(
    NAME recipeCalcTest_allGrain
    COMMAND brewtarget_tests recipeCalcTest_allGrain
 )
-
+ADD_TEST(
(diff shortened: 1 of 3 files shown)


rocketman768:
@mikfire does this make sense? I think it makes sense if and only if the following is a valid way to calculate the OG regardless of losses ([Testing.cpp:139](https://github.com/Brewtarget/brewtarget/blob/master/src/Testing.cpp#L139)).

    double plato =
       grain_kg
       * twoRow->yield_pct()/100.0
       * rec->efficiency_pct()/100.0
       / (rec->batchSize_l() * og) // Total wort mass in kg (not L)
       * 100; // Convert to percent

malavv:
This code looks fine and seems to be doing exactly what you meant.
I am not as knowledgeable when it comes to brewing formulas but it seems alright to me.
LGTM

Issue #47, opened 2015-06-03
Critical regression in several tools due to auto unit display


theophae:
The refractometer tool is broken because of the automatic conversion to "sg" gravity for all input fields. Because of that, it is impossible to correctly fill the input fields. For example, il you want to enter 15 Plato in the Original Plato field, it will be automatically replaced by 15.000 sg
I've seen a bug on the strike water calculator too. This time, the unit entered is not read, but it choose the smallest unit instead. To enter the weight of the grain, if I enter 5 kg it will change it to 5 mg. If I really want to put 5 kg I need to enter 5 000 000, then 5 kg is displayed.

@mikfire : This bug is coming from the BtLineEdit widget you've added. I think that this feature is very cool because it keeps the displayed field consistent. The problem is that for most of the tools, there are fields that needs to be filled with a specific unit, so we need to add something to allow forcing it. I'm not really a specialist of QtCreator, so I don't know yet how to implement it, so if you have something in mind please tell me ;) 

The 2.2.0 version mustn't be released with this bug.

theophae:
I've seen a bug on the strike water calculator too. This time, the unit entered is not read, but it choose the smallest unit instead. To enter the weight of the grain, if I enter 5 kg it will change it to 5 mg. If I really want to put 5 kg I need to enter 5 000 000, then 5 kg is displayed.

I confirm that this bug it critical because it may touch every part of the software.

mikfire:
I thought I had fixed that at some point. Assign this to me.

theophae:
How can I assign it to you ?

mikfire:
Beats me. As far as I've been able to see, only rocketman can assign tickets.

rocketman768:
Yeah, this part of github's interface sucks.

mikfire:
I think I can provide a per-field over ride. It will take me a few days to worm back through the unit system and make sure my idea will work. 

In the meantime, can somebody generate a list of the ui files (at least) that need to be fixed?

theophae:
These are the fields that have a problem:

OG Calculation help:
 - Plato
 
Refractometer Tools:
 - Original Plato
 - Current Plato

Equipement editor:
 - Mass (Problem with mass conversion)
 - Specific heat (always display "?")
 
Name Mash Editor:
 - Tun mass (Problem with mass conversion)
 
Strike water calculator:
 - Weight of grain (Problem with mass conversion)
 - Grain Weight (Problem with mass conversion)

[...]

PR #51, opened 2015-06-15
Bug #47 -- FSM weeps

mikfire:
Who knew that for the Q_ENUM to do its job, the enum had to be
declared inside the class definition? And without that, I couldn't do what I
wanted. So I did.

To move it inside the class means I needed full references
(Unit::unitDisplay, Unit::unitScale and Unit::unitType) everywhere they are
used. You know, I had wondered why that was the case. Anyway, that was about
60 files. Then I had to redo all of the signal/slot stuff. 10 more ui files
for that.

I know I keep promising not to do this. I wouldn't believe my promises any
more :)

Anyway. Bug 47 is almost done. I just need to update a few more files.


rocketman768:
@mikfire, with the other commits merged, should this be the end of bug #47?

mikfire:
I believe so.

Mik

PR #52, opened 2015-06-23
BtLineEdit changed to use the forced unit as default unit and allow e…

theophae:
…rasing field.

This is a little modification but it makes the user experience much more natural.
To me, it was very annoying to have to put the unit manually when editing a field which
has a forced unit. It is much more natural to just enter the value.
The other point is that it is now possible to erase a field. This was a personal frustration.
Now I can't see any more problem with this feature :)

### DIFF
--- a/src/BtLineEdit.cpp
+++ b/src/BtLineEdit.cpp
@@ -134,33 +134,40 @@ void BtLineEdit::lineChanged(Unit::unitDisplay oldUnit, Unit::unitScale oldScale
    if ( _property.isEmpty() )
       initializeProperties();
 
-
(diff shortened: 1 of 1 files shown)


mikfire:
Yeah. I do need to fix that, but I wasn't sure if it counted as an
enhancement or not.

Mik

theophae:
The impossibility to erase a field was not really a bug, but mixing the two problems, it leads to a bug.
When you wanted to erase the field "original Plato", it was equivalent to set to 0 without unit so it converted 0 sg to Plato.
This only way to avoid that was to write 0 P explicitly

PR #53, opened 2015-06-23
Change floating point format in beerXML to fit with the standard

theophae:
change the floating point format to better match beerXML v1.0

From the specification:
"- Floating Point - A floating point number, usually expressed in its simplest form with a decimal point as in "1.2", "0.004", etc...  Programs shall endeavor to store as many significant digits as possible to avoid truncating or losing small values.'

This should have no impact, but it should ensure that the exported beerXML files will be readable by every software that respects the standard.

### DIFF
--- a/src/BeerXMLElement.cpp
+++ b/src/BeerXMLElement.cpp
@@ -173,7 +173,7 @@ QString BeerXMLElement::text(bool val)
 
 QString BeerXMLElement::text(double val)
 {
-   return QString("%1").arg(val, 0, 'e', 5);
+   return QString("%1").arg(val, 0, 'f', 8);
 }
 
 QString BeerXMLElement::text(int val)


rocketman768:
## Long Story Short

I am OK with the fixed-point change, as long as we increase the number of digits from 5 to 8. Amend the commit, and I will merge.

## Some History

Exponent format has been in brewtarget since I first wrote it. It was a conscious decision so that the number of significant digits of each amount was the same regardless of the magnitude of the amount. The risk you run with fixed-point representation is having few or no significant digits for small numbers.

I communicated this to Brad Smith of Beersmith many years ago, and he agreed. I asked him whether or not the BeerXML standard should allow exponent format for floating point numbers, and he said yes. However, he never did anything.

Exponent format used to be very key to Brewtarget, since the xml output was not just for exporting, but rather the whole database was essentially a big set of xml files. With fixed-point format, the recipe numbers changed a little bit every time you opened and saved a recipe, especially with hops, since the amount is in kg, but in a recipe it is common to have very small amounts like 0.5 oz == 0.0141748 kg, which gets truncated and saved as 0.01417 kg == 0.499832 oz.

Since this is all stored properly in a database now, and the xml is only for output, I think the change is OK, but I would still like to increase the number of digits from 5 to 8 for more insurance against tiny amounts.

theophae:
Commit amended !

I personally completely agree with you that the exponent notation is a better choice, bug as we must follow the standard, we must change it.
I hope that 8 digit is enough, but we will be able to change in later if needed.

PR #56, opened 2015-06-29
Fixes wrong field display.

theophae:
BtLineEdit includes all the display unit management inside, all value to display must be passed to setText() in the default sytem unit. It is BtLineEdit who takes care of displaying the value in the right unit. Doing this way is ok, but we may find a better way of handling. You could create a class BtValue which would contains its value is default unit, and have all necessary methods to return the value in another unit. This is an idea, but I don't know if it is a good one.

It's really frustrating that every fixes regarding BtLineEdit seams to break something else every time.


theophae:
Please wait I say it's OK because I'm finding this issue in other tools. Its better to do only one merge for the same bug. Let me check as much as possible to find them all.
If you see one bug like this, tell me and I'll add the fix to my commit.

theophae:
Ok, now I think that there is no more bugging field (until the next one ><). But anyway this is correcting serious bugs.

PR #58, opened 2015-07-01
Add dynamic signal connection for yeast addition + ABV update

theophae:
I've previously added the function to update recipe after yeast change, but it took only the yeast already in the recipe at loading. I've forgotten to add the signal connection after yeast addition. I've also added the ABV recalc for recipe update after any yeast change. I hope not forgetting anything this time.

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -1845,10 +1845,18 @@ void Database::addToRecipe( Recipe* rec, Style* s, bool noCopy )
 // Why no connect here?
 void Database::addToRecipe( Recipe* rec, Yeast* y, bool noCopy )
 {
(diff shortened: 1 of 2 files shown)


rocketman768:
Agh, I should have caught this. Glad you did.

:+1:

theophae:
It was easy not to see it, because it quite rare to use more than one yeast. This should work well with my fix, but I think that it could be a good idea to trigger the recalculation of all the parameters that are affected by the changed value.
For example in this case: changing OG/FG => ABV must be recalculated.
Doing like this, the values will be recalculated only if needed.
I don't know the best way to do this, but I think that it all triggering action must be defined in the recipe constructor.

PR #59, opened 2015-07-03
Corrections of some translation parameters.

theophae:
Some text mustn't be translatable. For example, date format field sould not be changed.
This will make translators'life easier and avoid some bugs.

### DIFF
--- a/src/brewtarget.cpp
+++ b/src/brewtarget.cpp
@@ -1521,7 +1521,7 @@ QMenu* Brewtarget::setupMassMenu(QWidget* parent, Unit::unitDisplay unit, Unit::
          generateAction(sMenu, tr("Pounds"), Unit::scaleSmall, scale,qsgrp);
          break;
    }
-   sMenu->setTitle("Scale");
+   sMenu->setTitle(tr("Scale"));
    menu->addMenu(sMenu);
 
    return menu;
@@ -1552,7 +1552,7 @@ QMenu* Brewtarget::setupTimeMenu(QWidget* parent, Unit::unitScale scale)
    generateAction(sMenu, tr("Hours"),   Unit::scaleMedium, scale, qgrp);
    generateAction(sMenu, tr("Days"),    Unit::scaleLarge, scale, qgrp);
 
-   sMenu->setTitle("Scale");
+   sMenu->setTitle(tr("Scale"));
    menu->addMenu(sMenu);
 
    return menu;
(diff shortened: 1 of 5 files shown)


rocketman768:
Should we even have these date formats? For example, I think nobody except the US will understand the format MM/dd/yyyy. Should we just remove the explicit formats where possible and use [QLocalte::dateFormat()](http://doc.qt.io/qt-5/qlocale.html#dateFormat)?

theophae:
It seems that these values are set as default and changed by the format set in parameters.
I agree that it would be a good idea to use QLocale::dateFormat() instead. I have to say I hadn't thought a lot about it as my point was more to exclude all the strings that mustn't be translated.
I'm going to change if I can change it the way you want.

mikfire:
I would prefer to keep it as is. Yes, most in the US will use the
MM/dd/yyyy format. I actually use yyyy/mm/dd and want the choice.

brewtarget uses SI units and formats by default. That implied to me that we
should do that for dates as well, while allowing the user to choose what
they want.

As for making them to be translated, I did that intentionally. Add you
said, I doubt a German user will understand MM/dd/yyyy as an abbreviation
so I wanted them to be translated into the locale - appropriate
abbreviation.

Mik

theophae:
To me, all date field must follow the parameter set in Brewtarget options, so this value is only a default value.
I wanted to change that because it is very tricky for a translator to know how to translate it correctly.
In my case, I'd translated MM/dd/yyyy to MM/jj/aaaa because in french it is "mois" for "month", "jour" for "day" and "année" for year.
That translation is OK for the text field of option dialog, but not OK if the string was in fact a parameter of the QDateTimeEdit field. If this field is wrongly translated, the interface will be broken.
I think we need to check all the date fields to make then consistent.

rocketman768:
Let's split off the the changes to the translation flag for date strings to another pull request, because the other changes here look good-to-go now.

I would like to remove the month strings where possible (for example, in the options dialog, we should change the radio button text from "YYYY-dd-mm" to just "ISO Format"). We also need to make sure we use the selected option for the date everywhere we can (probably involving copying the code [here](https://github.com/Brewtarget/brewtarget/blob/master/src/PitchDialog.cpp#L63) to a new method in brewtarget.cpp).

theophae:
Personally I don't want to change the text "YYYY-dd-mm" to "ISO Format" because is much less clear. I personally didn't know what was the ISO format before, so the average user wouldn't know that.
I will split the pull request to discuss the date problem separately.

theophae:
This commit can be merge now. It contains no more change on date fields.

rocketman768:
:+1:

PR #60, opened 2015-07-03
Changes some checkboxes to radioButton when only one choice is possible.

theophae:
To me it much more logical and it was almost nothing to do.


rocketman768:
:+1:

PR #62, opened 2015-07-17
Add compilation flag to disable QtMultimedia.

theophae:
This allow to compile and execute Brewtarget without QtMultimedia.
It may allows developers to develop and test on the platforms that have a conflict on the Gstreamer version linked with QtMultimedia and QtWebKit.
The only drawback is that the timer can't play any sound.

This is a temporary fix for the bug #9


rocketman768:
:+1:

PR #66, opened 2015-07-31
Improve recipe export (text) & add BBCode export

theophae:
I changed almost all the behavior of the getTextFormat function in order to make it smaller and easier to maintain. I split this function in order to reuse all sub-function for BBCode extraction.
I change many thing to use the same unit and scales as in the main interface (it seems more logical).
Doing this I found a bug on the displayAmount function which was trashing the precision parameter in some cases.
I think that it can still be improved, but it is quite OK now.
I didn't give an implementation for the BBCode function as it would be considered as a new feature (even if I'd really like to see it included into 2.2.0), but it will be very easy to write know, as it will be base on the text format extract.
One other thing I did, is to clean the calories management. Before that it was sometime per 33cl and other times per 35cl.

@Brewtarget/developers : Please review and give me some feedbacks

### DIFF
--- a/src/BrewDayScrollWidget.cpp
+++ b/src/BrewDayScrollWidget.cpp
@@ -349,28 +349,29 @@ QString BrewDayScrollWidget::buildTitleTable(bool includeImage)
             .arg(tr("Boil Volume"))
             .arg(Brewtarget::displayAmount(recObs->boilVolume_l(), "tab_recipe", "boilVolume_l", Units::liters,2))
             .arg(tr("Preboil Gravity"))
-            .arg(Brewtarget::displayAmount(recObs->boilGrav(), "tab_recipe", "og", Units::sp_grav,0));
+            .arg(Brewtarget::displayAmount(recObs->boilGrav(), "tab_recipe", "og", Units::sp_grav, 3));
 
    // fourth row: Final volume and starting gravity
    body += QString("<tr><td class=\"left\">%1</td><td class=\"value\">%2</td><td class=\"right\">%3</td><td class=\"value\">%4</td></tr>")
             .arg(tr("Final Volume"))
             .arg(Brewtarget::displayAmount(recObs->finalVolume_l(), "tab_recipe", "finalVolume_l", Units::liters,2))
             .arg(tr("Starting Gravity"))
-            .arg(Brewtarget::displayAmount(recObs->og(), "tab_recipe", "og", Units::sp_grav, 0));
+            .arg(Brewtarget::displayAmount(recObs->og(), "tab_recipe", "og", Units::sp_grav, 3));
 
    // fifth row: IBU and Final gravity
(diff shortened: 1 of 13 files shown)


rocketman768:
Great! Reviewing...

rocketman768:
Looks like good code. Two suggestions for the text format:

1. "srm" should be capitalized to "SRM"
2. The units should be aligned vertically with each other somehow

I'm glad we can finally have BBCode at some point :smile:

theophae:
Regarding the point 1, I agree. I think that we must change several unit display:
sg -> SG
ebc -> EBC
srm -> SRM

I'd like to make unit translatable. The best example is "day" that stays in English in every languages.
But I think that theses changes should be done in another branch.

I also agree about the unit alignment, I try to do something about it.
Do you want me to add BBCode to this pull request ? Because right now it will be nothing to develop.

rocketman768:
Sure; go ahead with the BBCode.

rocketman768:
I just noticed the refractometer tools no longer allow fractional input. On the input side, all of the fractional parts of the field get rounded to a single digit. For example "7.87 P" becomes "8 P" and "1.045 sg" become "1 sg".

theophae:
I've notice some similar bugs. But I know why it appends.
I've corrected a bug that trashed the precision asked for the displayed value. So everywhere the precision is 0 it is now taken into account. I need to check every displayAmount call to see if the precision passed is correct. By the way, I think that a precision of 3 is too big in most of the case. It would be better to pass it to 2.
@mikfire : If I misunderstood something about the unit system please tell me.

theophae:
I've fixed the bug on gravity and color display. You can check if you see other bugs.

theophae:
I've added the BBCode export, but I don't know how to connect it to the interface. I thought of two solutions:
* To add an option to chose between plain text format or BBCode format using the current button
* To add a new menu entry for it

What do you prefer ?

rocketman768:
Gravity bug looks fixed.

I prefer another menu entry for BBCode export. It should probably be in the File->Recipe menu.

theophae:
Now you can export to BBCode from File->Recipe.
I've also improved instruction formatting for multiple line instructions.

I didn't change the alignment on the unit as rocketman768 asked because it is not trivial because the amount and its unit come directly together from displayAmount function.
If anyone see a simple way to do it, please tell me.

Enjoy BBCode ;)

rocketman768:
Thanks @theophae! If nobody else has any comments, go ahead and squash these commits.

theophae:
I've change the BBCode export to set tables' first line to bold (I wanted to do it at start but I've just figured out how to do it properly).
Squashed all commits so you can merge it.

theophae:
As it hasn't been merge yet, I've added a function to wrap too long instruction strings to keep most of the text below the 80 chars limit.
I think that we should rewords some of the instruction (as the "Add grain" instruction) which are not very clear. But it will be in another pull request.
Unless you ask me to change some points, I won't make any more change ;-)

PR #71, opened 2015-08-08
Fixes #65: bad ingredient amount behavior in non-US locales

### LINKED ISSUE #65: Wrong conversion from grams to kilograms in hops input
I've been using the .deb downloaded version so far, but today I tried the repository version. This bug happens on both branches (master and stable/2.2.0).

When trying to input an amount that has 1g precision in the hop amount field, Brewtarget automatically converts the input value to kilograms.

Multiples of 10 (10, 20, 30...) are accepted, anything else is converted to kilograms (1, 5, 7, 9, 15, 44, etc).

rocketman768:
The problem was that simple double properties of the ingredients
were being passed through the localization procedures in the
Brewtarget class when they should not have been.

### DIFF
--- a/src/fermentable.cpp
+++ b/src/fermentable.cpp
@@ -222,24 +222,20 @@ const QString Fermentable::typeStringTr() const
    return typesTr.at(type());
 }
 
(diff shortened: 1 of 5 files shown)


theophae:
+1

PR #74, opened 2015-08-23
[WIN] Correction of build issues.

malavv:
Trying to build Brewtarget on Windows revealed a few issues. I will try to find and eliminate them. Here are these included in this pull request.
* Inverted logic on Multimedia DLL inclusion.
* MINGW_BIN having a redundant bin path.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,7 @@ INCLUDE_DIRECTORIES(${Qt5LinguistTools_INCLUDE_DIRS})
 # Need to find SVG component to locate svg dlls later
 IF(WIN32)
    FIND_PACKAGE(Qt5Svg REQUIRED)
+   INCLUDE_DIRECTORIES(${Qt5Svg_INCLUDE_DIRS})
 ENDIF()
 
 # Fuckin Qt5 requires -fPIC if Qt5 itself was built with -fPIC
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -404,7 +404,7 @@ IF( WIN32 )
         ${QtQuick_location}
         ${QtPositioning_location}
    )
-   IF( NOT ${NO_QTMULTIMEDIA})
+   IF(${NO_QTMULTIMEDIA})
       SET( Qt_DLLs,
         ${Qt_DLLs}
         ${QtMultimedia_location}
```
rocketman768: This is adding QtMultimedia DLLs. I think we should be adding these DLLs unless NO_QTMULTIMEDIA is ON, right?

malavv: Thanks Phillip. You're completely right. I got bitten by the double negative. There is still an error because the original version does not seem to work but I'll revert that change.

malavv:
Hi Phillip and thanks again for looking at this. Here are the new modifications : 
* I reverted the double negative.
* I found the error which I was seeing (a trailing comma)
* I modified a comment about the bin folder to indicate the inclusion of the "bin".

malavv:
Ready for review and comments.

rocketman768:
:+1:

Go ahead and squash the commits together.

PR #76, opened 2015-08-30
Removed redundant namespace.

malavv:
- In brewtarget.h, there were many case of the use of Unit::Unit:: which might have
    been caused by a faulty find and replace.

  - G++ and Clang seemed ok with accepting this but not MSVS.


rocketman768:
:+1:

PR #80, opened 2015-09-10
Closes #48: more accurate ABV calculation

### LINKED ISSUE #48: consider alternate ABV calculation method
I could be misinformed, but it seems the ABV calculation method used in brewtarget is less accurate for higher gravity beers. For a large barleywine this accounts for 1%+ discrepancy in calculated ABV.

Using the standard method my OG/FG of 1.098 works out to about 9.9%
If I calculate using the "alternate" method which is reputed to have better accuracy on higher gravity beers I arrive at 11%.

My source on this information is http://www.brewersfriend.com/2011/06/16/alcohol-by-volume-calculator-updated/

btw, thanks for great software, and Cheers!

theophae:
Implementation of a new calculation method for ABV. It reduce de error on ABV calculation for strong beers. Under 5% of alcohol the result is quite similar to the former formula.
This closes the enhancement request #48.

### DIFF
--- a/src/recipe.cpp
+++ b/src/recipe.cpp
@@ -1412,10 +1412,9 @@ void Recipe::recalcABV_pct()
 {
    double ret;
 
-   // George Fix: Brewing Science and Practice, page 686.
-   // The multiplicative factor actually varies from
-   // 125 for weak beers to 135 for strong beers.
-   ret = 130*(_og_fermentable - _fg_fermentable);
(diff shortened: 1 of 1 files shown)


rocketman768:
Please amend the commit message to have a headline like

    Closes #48: more accurate ABV calculation

so that github closes issue 48 [automatically](https://help.github.com/articles/closing-issues-via-commit-messages/) when this gets merged.

PR #84, opened 2015-09-17
Closes #6: real-time ingredient search

### LINKED ISSUE #6: Real time search feature
From [launchpad](https://bugs.launchpad.net/brewtarget/+bug/1424161)

It would be very handy if the ingredient list windows had some sort of search field, allowing us to search for some text string, and having the list view update to show only the matching items.
For example, I could then type "cara" in the search field and it would show me also items named like "Simpsons - Caramalt".

dpettersson:
New pull request based on the correct branch/revision.

### DIFF
--- a/src/FermentableDialog.cpp
+++ b/src/FermentableDialog.cpp
@@ -32,6 +32,8 @@
 #include "MainWindow.h"
 #include "fermentable.h"
 
+
+
 FermentableDialog::FermentableDialog(MainWindow* parent) :
    QDialog(parent),
    mainWindow(parent),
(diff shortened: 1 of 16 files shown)


rocketman768:
This looks much easier than I though.

Could we implement a [QValidator](http://doc.qt.io/qt-5/qvalidator.html) to plug into [QLineEdit::setValidator](http://doc.qt.io/qt-5/qlineedit.html#setValidator) so the user has feedback on whether or not the filter is a valid regular expression? Otherwise, you can have an unnoticed syntax error that returns no results. I imagine the validator would be simple and just use [QRegularExpression::isValid](http://doc.qt.io/qt-5/qregularexpression.html#isValid).

dpettersson:
Il take a look at it (probably during the weekend).

theophae:
:+1:  Great! It was really one on the enhancement I really wanted :-)
Is it possible you add it for the other ingredients like hop and yeast ?

dpettersson:
I have started to wonder if we really need a custom validator for this. A faulty expression will not give any matching rows. The vast majority of people will search for the text that they want to find, and they will remove characters if the box goes blank. I will add the validator of you want it, but i can't really see how it would help us.

rocketman768:
I am concerned that 99% of the users won't know what a regular expression is. For example, they may expect a string without properly-matched braces to return a result, or they may search for something with a '+' with no results.

Another option is to do non-regex filtering and forget the validator.

dpettersson:
I changed to fixed string. A faulty string will not give any results. The regexp-part was a little over the top. :) 
Anything else that i need to change?

rocketman768:
Looks good for fermentables. Now just replicate to the other tables like @theophae suggests. You can also go ahead and squash the commits with the commit message headline `Closes #6: ...`

src/FermentableDialog.cpp
```diff
@@ -47,14 +49,16 @@ FermentableDialog::FermentableDialog(MainWindow* parent) :
    tableWidget->setModel(fermTableProxy);
    tableWidget->setSortingEnabled(true);
    tableWidget->sortByColumn( FERMNAMECOL, Qt::AscendingOrder );
+   tableWidget->verticalHeader()->setDefaultSectionSize(20);
```
theophae: You should remove this line. This should not be handled here but in FermentableTableModel.
And I quite sure that it is not a good idea to have a fixed value because it could leads to display bugs in some graphical environment.

dpettersson: I have removed the line as requested.
I will look into the problem with the resizing. Sorry for not beeing active but my internet connection broke down. It took the ISP several days to gt it back up.

dpettersson:
That should do it. Please review.
The problm with the row size disappeared when i disabled wordwrap.

theophae:
:+1: I don't know how you find that solution, but it works great!

rocketman768:
:+1: awesome

PR #85, opened 2015-09-22
Fixes #81: Improvement/weird check box

### LINKED ISSUE #81: Change weird checkbox in fermentable table
Since I use BT, I think the checkboxes that change when selected are very weird and unnatural.
For example, when the checkbox in unchecked, it indicate "steeped", but when I see it beside an unckecked checkbox, it seems to mean that the grain is not steeped.
I would be much clearer to have either a checkbox alone to indicate if the grain is mashed or a combo box with "Yes" or "No".
Same remark for the "Late Addition", it should use a combo box with "Normal" and "Late" as value.

If you think of other weird checkboxes like those, do not hesitate to add them.

I'd like to do this change, but I prefer to discuss what to do with you first.

theophae:
This is a first approach in order to change the weird checkbox used for Mashed option.
This commit aims to improve display avoiding manual resizing.
I need to find a way to hide one of the two parameters "Steeped" or "Not Mashed" regarding the fermentable type.

This branch fixes the bug #81

### DIFF
--- a/src/fermentable.h
+++ b/src/fermentable.h
@@ -49,8 +49,12 @@ public:
 
    //! \brief The type of Fermentable.
    enum Type {Grain, Sugar, Extract, Dry_Extract, Adjunct}; // NOTE: BeerXML expects a space for "Dry_Extract". We're screwed.
-   Q_ENUMS( TYPE )
-   
+   //! \brief The addition method.
+   enum AdditionMethod {Mashed, Steeped, Not_Mashed};
+   //! \brief The addition time.
+   enum AdditionTime {Normal, Late};
+   Q_ENUMS( Type AdditionMethod AdditionTime )
+
    virtual ~Fermentable() {}
    
    //! \brief The name.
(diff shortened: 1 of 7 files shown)


src/fermentable.h
```diff
@@ -49,62 +49,80 @@ class Fermentable : public BeerXMLElement
...
+   //! \brief The addition method.
+   enum AdditionMethod {Mashed, Steeped, Not_Mashed};
+   //! \brief The addition time.
+   enum AdditionTime {Normal, Late};
+   Q_ENUMS( Type AdditionMethod AdditionTime )
+
    virtual ~Fermentable() {}
    
    //! \brief The name.
-   Q_PROPERTY( QString name                  READ name                   WRITE setName                   NOTIFY changedName )
```
rocketman768: A lot of whitespace-only changes follow this line. Please revert.

theophae:
I'm ok to revert it, but I've added all these spaces to keep the original alignment.
Anyway, do not merge this branch until I find a proper solution to print only the right choices in the combobox.

rocketman768:
OK. Yes, please remove the whitespace changes. They make commits larger than necessary, harder to read, more likely to generate conflicts later, and impossible to `git blame` later, etc.

theophae:
I removed all unnecessary whitespaces.

rocketman768:
Thank you! Will merge when you give the go-ahead.

theophae:
I finally found a proper way to do what I wanted :-)
Please review before I squash all my commits.

rocketman768:
Sweet :+1:. Go ahead and squash.

PR #99, opened 2015-10-02
Argument parsing now using QCommandLineParser.

malavv:
Was looking for a way to add a CLI arguments and I found that the code in
main.cpp was using a hand build function for this with a todo to use the correct
"QCommandLineParser" functionalities coming from QT5.

I replaced the old code with the new module and this provides 2 new cli options.
-h for help and -v for version.

This modification was made in order to pave the way for a cli configurable configDir.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -20,61 +20,13 @@
  */
 
 #include <QApplication>
-#include <QStringList>
-#include <QHash>
+#include <QCommandLineParser>
 #include "config.h"
 #include "brewtarget.h"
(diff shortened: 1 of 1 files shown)


rocketman768:
Ahh, great. I love Qt5 :)

:+1:

malavv:
@rocketman768, thanks for the lgtm.

You will have to merge as I do not have write-access on this repo.

rocketman768:
Yup, I am the integration manager.

PR #102, opened 2015-10-03
Fixes #90: OG Correction Dialog auto unit conversion breaks the use case

### LINKED ISSUE #90: OG Correction Dialog auto unit conversion breaks the use case
Do the OG correction once. Now try to erase the SG/Temp fields and enter the Plato instead. This puts "0.000 sg" and "0.000 C" into the fields, and therefore doesn't let you do the Plato-based calculation.

f1oki:
Fixes OG Correction Dialog calculation from Plato

The change allows an empty LineEdit.

I also found that in the Plato conversion, the variable plato (supposed to be in P units) was in SG units (SI).

### DIFF
--- a/src/BtLineEdit.cpp
+++ b/src/BtLineEdit.cpp
@@ -134,6 +134,10 @@ void BtLineEdit::lineChanged(Unit::unitDisplay oldUnit, Unit::unitScale oldScale
    if ( _property.isEmpty() )
       initializeProperties();
 
+   if (text().isEmpty())
+   {
+      return;
+   }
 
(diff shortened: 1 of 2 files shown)


rocketman768:
Cherry-picked as [b5b0812](https://github.com/Brewtarget/brewtarget/commit/b5b081204d508cd551f2050969e95c8e7211d7e0) in master and [6b79d6a](https://github.com/Brewtarget/brewtarget/commit/6b79d6a241b3c16cbd1580489cba23452d4e8e95) in stable/2.2.0

PR #105, opened 2015-10-06
Fixes #98: Crash when trying to create a new folder

### LINKED ISSUE #98: Crash when trying to creat a new folder
I you try to create a folder in any of the different tabs, after entering the folder's name, it will crash. When I debug it, I find it bug at BtTreeModel.cpp line 483 ( beginInsertRows(parent,row,row); ) but I really don't know why.
This bug happens on windows 7 with the current master version (it is not a new bug, but I wasn't sure of my compilation chain at the time). On linux it works normally.

theophae:
I finally found the bug (quite by chance). It works great know !
I'm surprised myself that I fixed this bug ^^
This was the last bug to fix for milestone 2.2.0, so lets release !

I separated the changes in two commits in case you want me to revert the first one. But it seem always a good idea to avoid useless inclusion as it speeds up the compilation.

EDIT: Sadly, it crashes when inserting a recipe now ><
*#@$!*% stupid bug !

### DIFF
--- a/src/BtTreeModel.cpp
+++ b/src/BtTreeModel.cpp
@@ -480,7 +480,7 @@ bool BtTreeModel::insertRow(int row, const QModelIndex &parent, QObject* victim,
 
    bool success = true;
 
-   if(victim->objectName() == "BtFolder")
+   if(victimType == BtTreeItem::FOLDER)
       beginInsertRows(QModelIndex(),row,row);
    else
       beginInsertRows(parent,row,row);
--- a/src/BtTreeModel.cpp
+++ b/src/BtTreeModel.cpp
@@ -480,7 +480,10 @@ bool BtTreeModel::insertRow(int row, const QModelIndex &parent, QObject* victim,
 
    bool success = true;
 
-   beginInsertRows(QModelIndex(),row,row);
+   if(victim->objectName() == "BtFolder")
+      beginInsertRows(QModelIndex(),row,row);
+   else
+      beginInsertRows(parent,row,row);
    success = pItem->insertChildren(row,1,type);
    if ( victim && success ) 
    {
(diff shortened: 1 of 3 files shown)


theophae:
As it was working before my first try to fix the bug I deduce that a mix of both solution could work.
I don't know exactly how it works, but I think that when a folder is created a new list must be created, so we must differentiate tow cases.
If anyone have a better explanation I do want to hear it.

This time it should be OK

theophae:
Even if my change seems to work, I get a
```
QSortFilterProxyModel: invalid inserted rows reported by source model
```
:S

src/BtTreeModel.cpp
```diff
@@ -480,7 +480,10 @@ bool BtTreeModel::insertRow(int row, const QModelIndex &parent, QObject* victim,
 
    bool success = true;
 
-   beginInsertRows(parent,row,row);
+   if(victimType == BtTreeItem::FOLDER)
+      beginInsertRows(QModelIndex(),row,row);
```
rocketman768: `QModelIndex()` constructs an invalid index; [see here](http://doc.qt.io/qt-5/qmodelindex.html#QModelIndex). The usual way to create indices is [QAbstractItemModel::createIndex()](http://doc.qt.io/qt-5/qabstractitemmodel.html#createIndex).

@mikfire: any suggestion?

theophae:
After reading Qt manual about the view/model pattern, I don't understand how it can crash on this line. At the time it crashes `parent` is valid `QModelIndex` and `row` is valid too. It is really weird.

rocketman768:
If the parameters are good, and the behavior changes when you add a few
lines of code, then maybe we are stack smashing somewhere else and it just
happens to crash here?

theophae:
That's what I think, but it is very hard to debug as it is related to Qt Model/View internal behavior which uses several threads.
I've track the creation of a new folder all the way down from the button click to the line it crashes (only when compiled in debug mode). It is really driving me mad !

rocketman768:
Valgrind with the memcheck tool might be useful here, even when run in
Linux. It will tell you if there is bad memory behavior.

rocketman768:
This bug was closed with #114

PR #108, opened 2015-10-11
Convert recipe scaling tool into a wizard

rocketman768:
- Unify batch/efficiency scaling
 - Set the right equipment after scaling


theophae:
Great ! :+1: 
My only remark is about the wizard default size which is to small
scale-wizard-brewtarget-2 2 0

PR #114, opened 2015-11-09
Fixes #98 : Crash when trying to create a new folder.

### LINKED ISSUE #98: Crash when trying to creat a new folder

malavv:
- Modified model changing for item children insertion.
  - Also added lots of const modifier to function in the model.

Note:
It's not as clean as it could but the other stuff is all good practice and lots of const that should already been here.

### DIFF
--- a/src/BtTreeItem.cpp
+++ b/src/BtTreeItem.cpp
@@ -79,9 +79,6 @@ int BtTreeItem::type()
 
 int BtTreeItem::childCount() const
 {
-   if ( childItems.isEmpty() )
-      return 0;
-
    return childItems.count();
 }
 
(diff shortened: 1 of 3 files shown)


malavv:
@JulianVolodia, this change could better follow good *pull request* practice. However, change is small, easily understandable, and a clear improvement to *coding* practice.

Since a fix was required quickly and that it's easy to see that the rest of the modification are not causing side-effect, and that the previous code was dangerous for *DRY* and *const* reason I propose that we integrate as is. However, that is a decision that @rocketman768 needs to take.

malavv:
New version integrates replacement of the +row, and reverted to K&R bracket style.

rocketman768:
:+1:

Fantastic work @malavv! Verified that valgrind no longer complains :smiley:

PR #119, opened 2015-11-22
Fixes #93: Convert units/formulas options to combo boxes

### LINKED ISSUE #93: Convert units/formulas options to combo boxes
Instead of groups of radio buttons, use a form layout of combo boxes for the sake of standard UI design.

Daedalus12:
While I was at it, changed the IBU Adjustment inputs from LineEdit to DoubleSpinBox. I couldn't determine how they're used though, so wasn't sure what the range should be. If someone knows, I'll modify the .ui file.

Side note: this is my first commit to this project and I just learned Qt, so I'm wide open to feedback!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ build/*
 
 # Kate project
 .kateproject.d/
+.idea
--- a/src/OptionDialog.cpp
+++ b/src/OptionDialog.cpp
@@ -21,7 +21,6 @@
 
 #include "OptionDialog.h"
 #include "brewtarget.h"
-#include "UnitSystem.h"
 #include "UnitSystems.h"
 #include "USWeightUnitSystem.h"
 #include "SIWeightUnitSystem.h"
@@ -34,9 +33,7 @@
 #include "PlatoDensityUnitSystem.h"
 #include "SgDensityUnitSystem.h"
 #include "CelsiusTempUnitSystem.h"
-#include <QButtonGroup>
 #include <QMessageBox>
-#include <QFileInfo>
 #include <QFileDialog>
 #include "MainWindow.h"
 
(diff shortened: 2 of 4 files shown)


rocketman768:
Good job on your first Qt programming! This dialog is finally going to look normal :smiley: 

First, looks like the layout has been broken so that the dialog doesn't properly resize:

brokenlayout

Replace the original topmost layout and everything should be fine.

Second, I think we can get rid of the magic indices in the `switch()` by setting the combobox data programatically. Qt allows you to [attach arbitrary data](http://doc.qt.io/qt-4.8/qcombobox.html#addItem) to the combobox items that you can query. For example:

    colorFormulaComboBox->addItem(tr("Moshers Approximation"), QVariant(Brewtarget::MOSHER));

and later something like

    switch (colorFormulaComboBox->itemData(colorFormulaComboBox->currentIndex())) {
    case Brewtarget::MOSHER:
    ...
    }

Good job again, and welcome to the project.

Daedalus12:
Okay, think I've got both those resolved.

Only thing left is to set the value ranges should be for "First Wort" and "Mash Hop"; looks like they should be between 0 and 100?

rocketman768:
Yes, sounds good.

mikfire:
I would suggest higher at least for the fwh. The default for first wort
should be 110. It's used to calculate how much more ibu you get from a fwh
than a 60min addition.

Mik

rocketman768:
Oops, yeah. Thanks mik. Up to 200 then.

rocketman768:
Looks like `tab_2` has a broken layout.

brokenlayout1

Daedalus12:
Okay, should be all set now!

rocketman768:
One more change: remove redundant layouts and spacers. Once you do this, go ahead and [squash all the commits together](https://github.com/Brewtarget/brewtarget/wiki/Collaborating#pull-requests) and I'll merge.

Currently, layouts look like

brokenlayout2

They should look like

brokenlayout3

brokenlayout4

Daedalus12:
Sorry about that, still getting the hang of Qt Designer. 

I assumed you wanted me to squash and do a force-push on my fork's issue branch, let me know if I misunderstood.

rocketman768:
No problem. Yes, you are correct.

rocketman768:
:+1:

PR #121, opened 2015-11-27
Made "First Brewed Date" in tree use user format.

Daedalus12:
Moved formatted date method to brewtarget.cpp for reuse.


rocketman768:
Looks good. Since this is a bugfix, can you edit the pull request to use the `stable/2.3.0` branch as the target?

Daedalus12:
Can't edit it, opening a new one.

PR #123, opened 2015-12-02
Made "First Brewed Date" in tree use user format.

Daedalus12:
Moved formated date method to brewtarget.cpp for reuse.


rocketman768:
This was previously #121 

:+1:

PR #124, opened 2015-12-13
Refactored and extracted logging facilities.

malavv:
I had two problems with the current state of Brewtarget. First, the log
file was pretty useless when trying to debug error because  it's used
in too few places and does not store major events as to used when
looking at the user flow.

The second was that I am in thre process of trying to remove all OS
specific processing from Brewtarget.h and Brewtarget.cpp. I think this
class shouldn't care on which platform it runs and should also be way
clearer as it is one of the central classes.

Here is what was done.
  - Extracted all logging functionalities whilst keeping the same wrapper
    in Brewtarget.
  - Created a Log class to handle all logging.
  - Log is now using RAII on all stream and file descriptors.
  - Log is also now thread-safe.
  - Added INFO and DEBUG case for future inclusion in the application.
  - Log now stores a timestamp
  - Log now use a centralize template.
  - Log also calls a signal on all log entry.
  - The MainWindow now connects to this signal to update the statusBar.

### DIFF
--- a/src/Log.cpp
+++ b/src/Log.cpp
@@ -1,0 +1,92 @@
+/*
+ * Log.cpp is part of Brewtarget, and is Copyright the following
+ * authors 2009-2015
+ * - Maxime Lavigne <duguigne@gmail.com>
+ *
+ * Brewtarget is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Brewtarget is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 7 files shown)


rocketman768:
Awesome!

src/Log.h
```diff
@@ -0,0 +1,77 @@
...
+ * \class Log
+ *
+ * \brief Provides a proxy to an OS agnostic Log file.
+ */
+class Log : public QObject
+{
+   Q_OBJECT
+
+public:
+   Log(const QString defaultDir, bool isLoggingToStderr = true);
```
rocketman768: Is this class meant to be instantiated only in `class Brewtarget`? If so, we might want to make the constructor private, and add `Brewtarget` as a friend class to make it clear that you shouldn't create instances by yourself.

src/Log.cpp
```diff
@@ -0,0 +1,93 @@
...
+      warn(QString("Log is in a temporary directory: %1").arg(file.fileName()));
+      return;
+   }
+
+   warn(QString("Could not create a log file."));
+}
+
+Log::~Log() {
+   if( stream )
+      delete stream;
```
rocketman768: No need to check `stream`. The `delete` operator is safe on null pointers.

malavv:
Thanks Phillip,

Included all of your comments and rebased.

rocketman768:
:+1:

PR #126, opened 2015-12-14
App now has a visual hint of unsaved changes.

malavv:
A lot of the comments I was getting from users was that they did not know
about how Brewtarget didn't save anythin until manually asked to. Since
the application wasn't warning them, I think its fairly normal to wonder
why you were asked about this when closing the app.

Now when the dirty status changes on the DB, the statusBar reads "Unsaved
Changes"

### DIFF
--- a/brewtarget.qrc
+++ b/brewtarget.qrc
@@ -14,6 +14,7 @@
         <file alias="images/edit-copy.png">images/edit-copy.png</file>
         <file alias="images/exit.svg">images/exit.svg</file>
         <file alias="images/filesave.svg">images/filesave.svg</file>
+        <file alias="images/filesavedirty.svg">images/filesavedirty.svg</file>
         <file alias="images/flagBrazil.svg">images/flagBrazil.svg</file>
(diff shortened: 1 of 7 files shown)


rocketman768:
I forget we now live in a world of instant continuous saving...

rocketman768:
I like this. Since the save button is the way to commit changes, can we also make a version of the icon with a red dot (or something) in the corner which goes back to the original icon when the database is saved and not dirty?

malavv:
Sounds like a good idea. It should be pretty straightforward to implement using the new "is dirty changed slot".

Would you prefer me to put it in this pull request or do it separately?

rocketman768:
Cool. Do it in this PR.

malavv:
Per the comments of @rocketman768, this PR now adds a visual hint of unsaved changed in the save icon. 

Difference Save/Unsaved

dkoppenh:
A suggestion: make the "clean" icon grayed-out to indicate no save is necessary, and the "dirty" icon colored in to indicate the icon is active and a save is needed.
Regardless, this is a great improvement, thanks!

PR #129, opened 2015-12-20
Introduction of the user-dir cli option.

malavv:
This commit introduces the possibility of running brewtarget on a custom
database folder which will not impact your normal usage of the
Application.

This is intended for people with multiple version of Brewtarget and for
people helping test and debug the application.

  * Modification to the settings made while debugging the application no
    longuer interfere with normal settings.
  * Introduction of the --user-dir cli option which allows the caller to
    point to a directory containg the DB. This option will not store
    itself in the settings. Meaning running brewtarget again without param
    will revert to the known user data directory.
  * There were a lot of "file and directory arithmetics" implicated in
    this modification. Things like adding slash, double slash, backward
    slash, ... at the end of paths, and between path element. I stored the
    user data dir as a QDir instead of a QString to eliminate all these
    potential errors. It meant changing calls in a few file.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4,6 +4,7 @@
  * - A.J. Drobnich <aj.drobnich@gmail.com>
  * - Mik Firestone <mikfire@gmail.com>
  * - Philip Greggory Lee <rocketman768@gmail.com>
+ * - Maxime Lavigne <duguigne@gmail.com>
  *
  * Brewtarget is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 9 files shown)


rocketman768:
The concepts are good. The build bot and my local compiler complain about bad conversions in a few places.

malavv:
Thanks for pointing it out.

I'll review this whole commit and try to set the compiler in the correct mode to get all of this.

rocketman768:
What compiler do you you use?

malavv:
I'm using Mingw but it was not the cause of my problems. I forgot that code under a #define directive doesn't even get compiled. I've compiled under win, linux, and osx (using define) and their compiling now.

src/main.cpp
```diff
@@ -25,44 +26,61 @@
...
-void createBlankDb(const QString & optionValue);
+void importFromXml(const QString & filename);
+void createBlankDb(const QString & filename);
 
 int main(int argc, char **argv)
 {  
    QApplication app(argc, argv);
    app.setOrganizationName("brewtarget");
+#ifdef QT_DEBUG
+   app.setApplicationName("brewtarget-debug");
```
rocketman768: This is intended to use a separate set of options when built in debug mode, right? If so, can you put a comment to that effect here?

rocketman768:
:+1:

PR #131, opened 2016-01-02
Break lines in the notes of the brew day html output

carlesmu:
The Brew day pdf/html output don't keep the original line breaks.

I've added a style in the css (white-space: pre-wrap) that keep them.

### DIFF
--- a/css/brewday.css
+++ b/css/brewday.css
@@ -124,3 +124,8 @@ img
    text-align:right;
 }
 
+#customNote
+{
+   display:block;
+   white-space:pre-wrap;
+}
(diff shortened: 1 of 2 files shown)


rocketman768:
Thanks for the pull request. Can you post before and after images here?

carlesmu:
Hi,

I've attached the pdf files before and after the change.

[brewday.after.pdf](https://github.com/Brewtarget/brewtarget/files/76741/brewday.after.pdf)
[brewday.before.pdf](https://github.com/Brewtarget/brewtarget/files/76742/brewday.before.pdf)

rocketman768:
:+1:

Issue #142, opened 2016-01-22
[master] core dump on creating new database


mikfire:
This is on the master branch. If you delete your ~/.config/brewtarget directory -- to look like a fresh install -- and run brewtarget, it dumps core.

      Program received signal SIGSEGV, Segmentation fault.
       0x00007ffff5a7101d in QLocalePrivate::dateTimeToString(QString const&, QDateTime const&, QDate const&, QTime const&, QLocale const*) const () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
      (gdb) bt
      #0  0x00007ffff5a7101d in QLocalePrivate::dateTimeToString(QString const&, QDateTime const&, QDate const&, QTime const&, QLocale const*) const () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
      #1  0x00007ffff5a727d2 in QLocale::toString(QTime const&, QString const&) const () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
      #2  0x00007ffff5a4fb8f in QTime::toString(QString const&) const () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
      #3  0x00000000005c669b in Log::doLog (this=0x9ddc20 <Brewtarget::log>, lt=Log::LogType_WARNING, message=...) at /home/mik/brewtarget/mik/src/Log.cpp:86
      #4  0x00000000005c6525 in Log::warn (this=0x9ddc20 <Brewtarget::log>, message=...) at /home/mik/brewtarget/mik/src/Log.cpp:75
      #5  0x0000000000501d7c in Brewtarget::logW (message=...) at /home/mik/brewtarget/mik/src/brewtarget.cpp:995
      #6  0x00000000004fc76b in Brewtarget::getConfigDir (success=0x0) at /home/mik/brewtarget/mik/src/brewtarget.cpp:363
      #7  0x0000000000507bea in __static_initialization_and_destruction_0 (__initialize_p=1, __priority=65535  at /home/mik/brewtarget/mik/src/brewtarget.cpp:88
      #8  0x0000000000507cd1 in _GLOBAL__sub_I_brewtarget.cpp(void) () at /home/mik/brewtarget/mik/src/brewtarget.cpp:1585
      #9  0x00000000006bb41d in __libc_csu_init ()
[...]

PR #143, opened 2016-01-23
Closes issue #142

mikfire:
This was an order of operations thing. During initialization of Brewtarget,
getConfigDir() was called before the log object was initialized. This caused
problems.

I simply re-ordered the lines to make certain the Log object was initialized
first.


mikfire:
Nope. Turns out I removed a log entry without realizing it. Reordering things doesn't work, but this has something to do with the log object being created before the directory exists

PR #146, opened 2016-01-23
Check for Visual Studio in pstdint.h

danielm5:
I've modified pstdint.h to check for VS2012 and up which have their own version of stdint.h, similarly to what's already being done for GCC. Without this change VS2015 report redefinition errors and the build fails.

### DIFF
--- a/src/pstdint.h
+++ b/src/pstdint.h
@@ -193,7 +193,11 @@
  *  do nothing else.  On the Mac OS X version of gcc this is _STDINT_H_.
  */
 
-#if ((defined(__STDC__) && __STDC__ && __STDC_VERSION__ >= 199901L) || (defined (__WATCOMC__) && (defined (_STDINT_H_INCLUDED) || __WATCOMC__ >= 1250)) || (defined(__GNUC__) && (defined(_STDINT_H) || defined(_STDINT_H_) || defined (__UINT_FAST64_TYPE__)) )) && !defined (_PSTDINT_H_INCLUDED)
+#if ((defined(__STDC__) && __STDC__ && __STDC_VERSION__ >= 199901L) \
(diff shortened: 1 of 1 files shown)


src/pstdint.h
```diff
@@ -193,7 +193,11 @@
  *  do nothing else.  On the Mac OS X version of gcc this is _STDINT_H_.
  */
 
-#if ((defined(__STDC__) && __STDC__ && __STDC_VERSION__ >= 199901L) || (defined (__WATCOMC__) && (defined (_STDINT_H_INCLUDED) || __WATCOMC__ >= 1250)) || (defined(__GNUC__) && (defined(_STDINT_H) || defined(_STDINT_H_) || defined (__UINT_FAST64_TYPE__)) )) && !defined (_PSTDINT_H_INCLUDED)
+#if ((defined(__STDC__) && __STDC__ && __STDC_VERSION__ >= 199901L) \
+      || (defined (__WATCOMC__) && (defined (_STDINT_H_INCLUDED) || __WATCOMC__ >= 1250)) \
+      || (defined(__GNUC__) && (defined(_STDINT_H) || defined(_STDINT_H_) || defined (__UINT_FAST64_TYPE__)) ) \
+      || (defined(_MSC_VER) && (_MSC_VER>=1700 || defined(_STDINT))) \
```
rocketman768: Thanks very much! For history, is there a document you can link to indicating VS2012 and up have their own definitions?

danielm5: I've found it on the documentation:
VS2012: https://msdn.microsoft.com/en-us/library/hh874765%28v=vs.110%29.aspx
VS2013: https://msdn.microsoft.com/en-us/library/hh874765%28v=vs.120%29.aspx
VS2015: https://msdn.microsoft.com/en-us/library/hh874765%28v=vs.140%29.aspx

rocketman768:
:+1:

PR #147, opened 2016-01-23
Fixed bad static initialization of config directory

danielm5:
Fixed the warning 'QCoreApplication::applicationDirPath: Please instantiate the QApplication object first' which appears on Windows builds. It also makes sure an empty QApplication pointer isn't used.

### DIFF
--- a/src/brewtarget.cpp
+++ b/src/brewtarget.cpp
@@ -85,7 +85,7 @@ QFile Brewtarget::pidFile;
 QDateTime Brewtarget::lastDbMergeRequest = QDateTime::fromString("1986-02-24T06:00:00", Qt::ISODate);
 
 QString Brewtarget::currentLanguage = "en";
-QDir Brewtarget::userDataDir = getConfigDir();
+QDir Brewtarget::userDataDir = QString();
 
 bool Brewtarget::checkVersion = true;
 Log Brewtarget::log(true);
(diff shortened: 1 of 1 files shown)


src/brewtarget.cpp
```diff
@@ -388,10 +388,16 @@ const QDir Brewtarget::getConfigDir(bool *success)
    return dir.absolutePath() + "/";
 
 #elif defined(Q_OS_WIN) // Windows OS.
+   if (!qApp)
+   {  //QApplication instance doesn't exist yet
+      if (success)
+         *success = false;
+      return QString();
```
rocketman768: This should never happen if the code is correct, so add an error log message here.

danielm5: It actually happens everytime because of the static member: 
Line 88: QDir Brewtarget::userDataDir = getConfigDir();
which is initialized before calling main(), i.e. before creating the QApplication.

rocketman768:
Ooh, ouch. Then either this function needs to be reworked so that we can use it to initialize the static member, or we need to explicitly initialize it at an appropriate place in `Brewtarget::initialize()`.

danielm5:
On Windows the static variable Brewtarget::userDataDir is always initialized to an empty QString() because QCoreApplication::applicationDirPath() returns empty (and reports the warning) if called before QApplication is created. The software still works as intended, as far as I can tell, because later getConfigDir() is called again by ensureDirectoriesExist() at line 137. My patch only makes this clear and unambiguous. This problem is only for Windows builds. I think adding userDataDir = getConfigDir() in initialize is also a good idea. Would you prefer me doing that instead?

rocketman768:
> I think adding userDataDir = getConfigDir() in initialize is also a good idea. Would you prefer me doing that instead?

Yes, that would be good.

danielm5:
The static variable userDataDir was already correctly set by Brewtarget::initialize(). I just removed the call to getConfigDir() when the static variable is created to avoid that call before QApplication is created. The warning is removed and not messing with empty application pointers.

rocketman768:
:+1:

Issue #73, opened 2015-08-14
Two Running Brewtargets Resets Database


eyousey:
I see there is a previous bug for this (#1109597) on lauchpad, but it is still happening.

I accidentally ran two versions (back-to-back) and now my database was reset back to the default. It seems you put in a lock file, but maybe in my case they both launched before the lock file was created.

I did not alter any recipe info when I had both (accidentally) open, I was just viewing. When asked to save my changes, I said, "No."

This being KDE 4.14.3 on FreeBSD 10.1 (64bit) I closed the "<2>" version first.

Can a check be added to only allow one client to connect to the database at once?

As a an additional level of safety, make a backup copy of the database at each launch and shutdown with a unique time stamp appended to the file name. Keeping maybe 10 or so backups in rotation. Potentially have this number of backups as a user adjustable number.

Thanks.

rocketman768:
Race condition between [brewtarget.cpp:478](https://github.com/Brewtarget/brewtarget/blob/master/src/brewtarget.cpp#L478) and [brewtarget.cpp:486](https://github.com/Brewtarget/brewtarget/blob/master/src/brewtarget.cpp#L486).

Desired fix is to replace the pid file logic with [QSharedMemory](http://doc.qt.io/qt-5/qsharedmemory.html) communication.

malavv:
Thanks @rocketman768 for finding the source of the issue.

**Considerations**
* It seems that whatever the fix is needs to be multi-platform.
* The two only things that are shared by application instance are Settings and Database.
* QSetting is reentrant and can be accessed by multiple instances.
* At this point, all instances will be running the same DB (due to user-data-dir in settings), but it might not be the case soon.

**Conclusion**
* Could we check for multiple access to the same DB instead? Since this is were the problems will come from?

mikfire:
This will be fixed soon. I've got a branch that seems to get the work done https://github.com/mikfire/brewtarget/tree/bugs/twobt but I need to work on it a bit more.

mikfire:
Clsoed with merge request #260

PR #151, opened 2016-01-26
PostgreSQL backend

mikfire:
This is NOT ready for a full merge into master, but I wanted to raise awareness and I need some help. 

This branch enables brewtarget to use PostgreSQL as its backend database. As it stands, I've implemented about 90% of what we need. I've added the configuration screens, I've added the dialogs, I've made DataSchemaHelper work, I've refactored that monstrosity into something a bit more usable, I've added functionality to import from SQLite to PostgreSQL.

Why? In a very simple word, mobile. I tried for months to get the SQLite file to sync between my mobile device and my desktop and failed. I started work on this branch on Thursday, and have enough of a working prototype to publish. If we get the data off the device, mobile becomes trivial.

Directions for setting up postgreSQL can be found in data/POSTGRES.markdown. The file also explains some of the reasons behind my doing what I did.

I need help with:
* the ability to merge an updated SQLite database into an existing PostgreSQL database;
* the interface is still very rough around the edges -- the configuration screen is way too chatty for example;
* the import is really fragile. I've found some glaring inconsistencies in the sqlite database, like fields with a null name that are defined as non-null, like fields specified as varchar(256) with 300+ characters in them, etc. Right now, those break the import. I need to know if those are simply artefacts in my database or if they are widespread;
* the import works in one direction: SQLite -> PostgreSQL. Do we need to do the reverse? I think I can, and likely will;
* there are some weird bits in the code -- notice what I had to do to unload and why;
* should we subclass Database and DatabaseSchemaHelper? Should we create a another low-level class to handle the differences between SQLite and other dbs (the booleans killed me, and don't even start with timedate stamps).

Things I haven't tried:
* honestly, I haven't had the courage to actually create anything new. That will happen tomorrow.

Things to consider:
* Any backups have to be done from postgresql, not brewtarget;
* No more backup databases, which is a plus and a minus;
* I still suspect I should have probably subclassed *something*, and I am afraid adding a third RDBMS solution (oracle, db2, mysql, etc) will really break things. So far, I tried to make it as extensible as I could, but ....


mikfire:
I think I may have left some extra commas in the table definitions in DatabaseSchemaHelper::create() for the id columns. I will double check and fix as required. 

I also need to fix the instructions -- the perl script was translated into Database::convertSqliteToPostgress() and should happen automatically.

mikfire:
I am saying we support only postgres 9.5 and later at the moment. I need that version for the upsert functionality. I can change that, but we lose some easy function for the inventory tables.

Right now, I copy the data from the old DB to the new DB on startup. It occurs to me that there is no reason I couldn't do that as soon as the options are set. Any preferences? I am sort of leaning towards sooner than later.

src/brewtarget.h
```diff
@@ -144,6 +145,14 @@ class Brewtarget : public QObject
 
    };
 
+   //! \brief Supported databases. I am not 100% sure I'm digging this
+   //  solution, but this is more extensible than what I was doing previously
+   enum DBTypes {
```
rocketman768: `DBType`

src/brewtarget.h
```diff
@@ -144,6 +145,14 @@ class Brewtarget : public QObject
...
+      NODB = -1,  // seems a popular choice with the cool enums
+      SQLITE,     // compact, fast and a little loose
+      PGSQL       // big, powerful, uptight and a little stodgy
```
rocketman768: Prefix these with `DBTYPE_`

src/database.h
```diff
@@ -101,20 +101,26 @@ class Database : public QObject
...
    static bool createBlank(QString const& filename);
-   
+
    //! backs up database to 'dir' in chosen directory
    static bool backupToDir(QString dir);
 
    //! \brief Reverts database to that of chosen file.
    static bool restoreFromFile(QString newDbFileStr);
 
+   static bool testConnection( Brewtarget::DBTypes testDb, QString const& hostname,
```
rocketman768: Rename `isConnectionGood()` or something that reflects what the return value `true` means.

src/DatabaseSchemaHelper.h
```diff
@@ -344,4 +352,52 @@ class DatabaseSchemaHelper
    
    //! \brief Current schema version of the given database
    static int currentVersion(QSqlDatabase db = QSqlDatabase());
+   static QString FOREIGNKEY( QString const& column, QString const& foreignTable );
+   static bool create_childTable( QSqlQuery q, QString const& tableName, QString const& foreignTable);
```
rocketman768: Stick with camel case.

rocketman768:
So, the changes to DatabaseSchemaHelper.cpp are basically a rewrite of the whole file. Can you summarize to help me try to reread and reunderstand the whole thing?

mikfire:
Not so much a rewrite, as it was a reorg with enhanced error reporting. The create() method was 600+ lines with no error reporting. If something went wrong, you had no idea what and it was a horrible time finding it. Please believe me, because it did and I didn't and it was.

In defense of my own sanity, I split creating the tables into 6 pieces: core (aka, BeerXMLElements), bt_* tables, *_in_recipe tables, triggers, children tables and inventory tables. 

Each core table got its own method for creating it. They are too unique to bundle up. The other tables were more repetitive and got one method that took the correct parameters to create the right table. The triggers are just annoying and have to be handled uniquely. Every create call now throws an error message if something fails. I would honestly prefer to stop all processing when an error is found, but I haven't had the patience to do that yet.

I also had to defer some of the strings for a late evaluation. I don't know some values at initialization time -- mostly the database type --  so they had to be set when create() is called. It got more complicated when I moved the database transfer from post to pre. 

migrateNext() just got split into its logical subcomponents. Since I was already there, the way to split it up was obvious and the size of migrateNext() is a strictly increasing function, I thought I should take the opportunity to set that right. I really do not like any method longer than 50 lines. It sometimes cannot be avoided, but that doesn't mean we shouldn't try. One day, I will get annoyed and split up MainWindow(), and it will be a very satisfying 3 days of janitorial work .

Is there any chance to argue you out of the DBTYPE_ prefix? I can put a few forward, but if there's no chance I will not waste my time. I will fix the rest.

mikfire:
This last one gets real close to done. I need to just make sure I haven't broken anything when using SQLite, and I really would like to get the postgresql -> sqlite conversion working.

I need a lot more error checking all through the code. Every place we touch the database needs to make sure the operation succeeded; networks can fail at any time and we should catch that.

rocketman768:
This is something Brewtarget needs for its future, so thank you for this.

I read most of it. It is huge though, so it's hard for me to offer any substantive comments. I'm sure I have failed to notice some bugs, so all I can ask is: does it work and what validation did you do? We will obviously need a good test plan to make sure there is no path to corrupt user data now that the backup db is gone.

Thanks for modularizing the schema helper. I agree with you: shorter readable functions are better.

Do we need to subclass Database/DatabaseSchemaHelper? That is not clear to me. If we ever decide that support of a very different database type is necessary, then it should become clear at that point. For now, meh. Maybe others' workflow is different, but I always write first and refactor later when I find myself copying too much code.

Is this still a Work In Progress? What else do you need?

mikfire:
The validation question is really hard to answer. I validated with my
*real* sqlite database. Not the default one, but the one that has my
brewing data since my first use of brewtarget around v1.2 days. I know it
works on my data. Don't know if it will work on yours. I am not usually a
good test subject, but my database is as dirty as it gets -- I've lost
track of the number of half baked ideas that have made an absolute mess of
my db.

The process is non-destructive. If it doesn't work, you just need to delete
some values from your brewtarget.conf file and it all reverts to sqlite and
no damage done. I've lost track of the number of times I've practised this
and it works. I've scattered enough log statements that if it does go
wrong, I can likely figure out what the malfunction was from those logs.

What I need.
1. I would really like somebody else to try this. It's non-destructive. Set
up a pgsql database and give it a run. Let me know what doesn't work. Take
me one step better, and do something on AWS.
2. I need to make a method for converting from pgsql->sqlite.  In my
scheming for mobile, I have a use case for easily converting between the
networked and non-networked dbs.
3. Error handling and transactions. When your db is a local file, you can
make any number of assumptions about things working. When its a networked
system, you can't assume anything worked because it can all fail. I have
been looking at this weekend, and it is a very hard problem.
4. Merging. Still haven't tested, still gives me sweats. This is closely
related to item 2.

Having said that. I brewed with the new code today, using pgsql as my
backing store. It worked without a hitch. I ran it side-by-side with what I
use for my real data and it all seems to have worked. I am likely going to
transition to the online version very soon. Once there, mobile is only a
build chain away.

Mik

On Sun, Jan 31, 2016 at 3:58 PM, Philip Lee <notifications@github.com>
wrote:

> This is something Brewtarget needs for its future, so thank you for this.
>
> I read most of it. It is huge though, so it's hard for me to offer any
> substantive comments. I'm sure I have failed to notice some bugs, so all I
> can ask is: does it work and what validation did you do? We will obviously
> need a good test plan to make sure there is no path to corrupt user data
> now that the backup db is gone.
>
> Thanks for modularizing the schema helper. I agree with you: shorter
> readable functions are better.
>
> Do we need to subclass Database/DatabaseSchemaHelper? That is not clear to
> me. If we ever decide that support of a very different database type is
> necessary, then it should become clear at that point. For now, meh. Maybe
> others' workflow is different, but I always write first and refactor later
> when I find myself copying too much code.
>
> Is this still a Work In Progress? What else do you need?
>
> —
> <https://github.com/Brewtarget/brewtarget/pull/151#issuecomment-177609046>
> .
>



-- 

rocketman768:
@Brewtarget/developers ... HEAR ME :smile:

> What I need.
1. I would really like somebody else to try this. It's non-destructive. Set
up a pgsql database and give it a run. Let me know what doesn't work. Take
me one step better, and do something on AWS.
2. I need to make a method for converting from pgsql->sqlite.  In my
scheming for mobile, I have a use case for easily converting between the
networked and non-networked dbs.
3. Error handling and transactions. When your db is a local file, you can
make any number of assumptions about things working. When its a networked
system, you can't assume anything worked because it can all fail. I have
been looking at this weekend, and it is a very hard problem.
4. Merging. Still haven't tested, still gives me sweats. This is closely
related to item 2.

rocketman768:
I'll play with my AWS account this week if it's easy enough.

malavv:
Hi @mikfire, item number 3 is important for another reason. It would fix issue #73. We could even remove the check (only done on Linux) about concurrent systems.

mikfire:
Ugh. You may want to avoid ec90bec for a bit. I tried to avoid do it, but I simply cannot clean things up without reworking them a little.

EvansMike:
Hi mikfire
I've just set up a local db with LInux and psql (9.2.1) 
On trying to delete a brew I get.
[14:09:28.436] ERROR : SetterCommand::redo: ERROR:  parameter $1 of type integer cannot be coerced to the expected type boolean
HINT:  You will need to rewrite or cast the expression.
(42804) QPSQL: Unable to create query.
   "UPDATE brewnote SET deleted=:value WHERE id=0"

mikfire:
Postgres 9.2.1 won't work. The upsert functionality I use exists in 9.5
only. And that error tells me I missed a boolean somewhere.

Thanks!

On Wed, Feb 3, 2016 at 9:14 AM, Mike Evans <notifications@github.com> wrote:

> Hi mikfire
> I've just set up a local db with LInux and psql (9.2.1)
> On trying to delete a brew I get.
> [14:09:28.436] ERROR : SetterCommand::redo: ERROR: parameter $1 of type
> integer cannot be coerced to the expected type boolean
> HINT: You will need to rewrite or cast the expression.
> (42804) QPSQL: Unable to create query.
> "UPDATE brewnote SET deleted=:value WHERE id=0"
>
> —
> <https://github.com/Brewtarget/brewtarget/pull/151#issuecomment-179258933>
> .
>



-- 

mikfire:
I found the boolean :) I think the last checkin gets back to basic functionality. I also fixed issue #164 and went with a straight, true delete. That took a while to understand and I don't think I even made use of the return codes I added. Tomorrow

mikfire:
I've brewed again with the new code. So far, I am ultra pleased with where it is going. Still needs a lot of work to make sure the error handling is handled properly :) Lots of places don't expect a null return, and bad things start happening. I will fix as I go through.

rocketman768:
> Postgres 9.2.1 won't work. The upsert functionality I use exists in 9.5
only.

Ah poop. AWS only has 9.4.5 support and plans to [support 9.5.1](https://forums.aws.amazon.com/thread.jspa?messageID=699462) when it arrives.

mikfire:
We really need to spend some time untangling the cross-class calls. I don't think any of it actually required, and it is making the transactional boundaries hard to figure out.

mikfire:
Ok. I have a problem I need others helping on. I am trying to test error cases, and I keep getting core dumps.

To reproduce, build this branch, start brewtarget and then stop the postgresql database. Bad things will happen, and core will be dumped. The problem is I can't figure out why. Here is what I think I know:
1. Stopping the database causes the postgresql driver to emit some signal (I think)
2. Qt catches that signal and starts refreshing everything, starting with the fermentable table
3. QSortFilterProxyModel->data(fermnamecol) is called, which calls fermentabletablemodel->data(fermnamecol), which calls database->get(name)
4. Database->get returns empty, fermentabletablemodel->data return empty
5. QSortFilterProxyModel dumps core.

I can catch the error in step 4, but I can't seem to do anything to stop step 5. Is it okay to throw a warning message at step 4 and simply close BT? If it isn't, can anybody else figure out why this is coring?

theophae:
I wanted to help by doing some test, but I really don't know how to set up the PostgreSQL DB. I've installed postgres-9.5 on my Debian and I've fought a bit against the basic configuration. Now I succeeded to create user with the necessary right to create a DB, and I've create an empty DB. But I don't know what to do next.

EvansMike:
I've not fixed the crash but I may have fixed another error I see.
```
[12:07:38.644] ERROR : bool Database::getElements(QList<T*>&, QString, Brewtarget::DBTable, QHash<int, T*>, QString) [with T = MashStep] could not execute query: SELECT id as id FROM mashstep WHERE mash_id = 51 AND deleted = 0 : ERROR:  operator does not exist: boolean = integer
LINE 1: ...CT id as id FROM mashstep WHERE mash_id = 51 AND deleted = 0
                                                                    ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
```
Apologies for the inline patch but Firefox keeps crashing when I attach it:

diff --git src/brewtarget.cpp src/brewtarget.cpp
index 13b7584..910271f 100644
--- src/brewtarget.cpp
+++ src/brewtarget.cpp
@@ -542,7 +542,7 @@ QString Brewtarget::dbFalse(Brewtarget::DBTypes type)
    QString retval;
 
    if ( whichDb == Brewtarget::NODB )
-      whichDb = (Brewtarget::DBTypes)option("dbType", Brewtarget::SQLITE).toInt();
+      whichDb = (Brewtarget::DBTypes)option("dbType", Brewtarget::dbType()).toInt();
 
    switch( whichDb ) {
       case SQLITE:

mikfire:
You are almost done. Just build this branch, then go to tools->options and
go to the database tab. Change the dbtype from the drop down, fill in the
information, test the connection, and follow the prompts.

I did write a quick howto in dev-doc/POSTGRES.markdown that may be of some
help.

Mik

theophae:
`I did write a quick howto in dev-doc/POSTGRES.markdown `

It seems that you've never committed it ^^

mikfire:
Oops. Color me embarrassed. Doesn't do me any good to write docs if nobody
side can read them.

I'll get that checked in tonight.

Mik

mikfire:
@EvansMike what were you doing when that error was generated? It implies there is one more place where I missed a boolean being misused, which will depress me but not surprise me. That your patch fixed the problem is more surprising.

EvansMike:
Hi mikfire
I was testing your postgres code to help with the crash, but as to **exactly** what I as doing? I forget. Glad to be of help though.
Mike

mikfire:
Right. I think the core is done. Anybody brave enough to merge this?

rocketman768:
Done. A few months remain for further testing and fixups.

PR #156, opened 2016-01-28
Closes #155: can't delete a style from the tree

### LINKED ISSUE #155: Can't delete a style from the tree
You can create and modify, but can't delete. This will be fixed directly.

mikfire:
Just forgot a case in my switch statement. That's been cleaned up, along with
a slight change to the error messages.


rocketman768:
:+1:

PR #157, opened 2016-01-28
Closes #153: improve equipment editor

### LINKED ISSUE #153: Improve Equipment Editor GUI
1. Batch size should be above Pre-boil volume as this is input first by the user.
2. Auto re-calculate Pre-boil volume whenever Batch size is update and if checkbox is checked. Currently this does not happen.
3 Don't allow manual update of Pre-boil volume if checkbox is checked.

EvansMike:
Changes layout and enables automatic computation of Pre-boil volume.

### DIFF
--- a/src/EquipmentEditor.cpp
+++ b/src/EquipmentEditor.cpp
@@ -75,7 +75,8 @@ EquipmentEditor::EquipmentEditor(QWidget* parent, bool singleEquipEditor)
    connect(lineEdit_evaporationRate,SIGNAL(textModified()),this,SLOT(updateCheckboxRecord()));
    connect(lineEdit_topUpWater,SIGNAL(textModified()),this,SLOT(updateCheckboxRecord()));
    connect(lineEdit_trubChillerLoss,SIGNAL(textModified()),this,SLOT(updateCheckboxRecord()));
-
+   connect(lineEdit_batchSize, SIGNAL( editingFinished() ), this,SLOT(updateCheckboxRecord()));
+                     
    // Set up the buttons
    connect( pushButton_save, SIGNAL( clicked() ), this, SLOT( save() ) );
    connect( pushButton_new, SIGNAL( clicked() ), this, SLOT( newEquipment() ) );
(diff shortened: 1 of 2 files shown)


EvansMike:
Based on master.

src/EquipmentEditor.cpp
```diff
@@ -730,7 +737,9 @@ void EquipmentEditor::updateCheckboxRecord()
    {
       double bar = calcBatchSize();
       lineEdit_boilSize->setText(bar);
+      lineEdit_boilSize->setReadOnly(true);
    }
+   else lineEdit_boilSize->setReadOnly(false);
```
rocketman768: These should be `setEnabled(false/true)` so that they are visually grayed out.

EvansMike: Personally I disagree. A disabled EditText makes it look as if the value is not important in this context. To quote from a MSDN forum post.

"Use ReadOnly when you have data that you want the user to see and copy, but not modify. Use a disabled textbox, when the data you are displaying is not applicable in for the current state of a dialog or window."

However, I will change it if you insist.

EvansMike: On Thu, 28 Jan 2016 19:30:16 -0800
Philip Lee <notifications@github.com> wrote:

> >     }
> > +   else lineEdit_boilSize->setReadOnly(false);
> 
> These should be `setEnabled(false/true)` so that they are visually grayed out.
> 


Personally I disagree. A disabled EditText makes it look as if the value is not important in this context. To quote from a MSDN forum post.

"Use ReadOnly when you have data that you want the user to see and copy, but not modify. Use a disabled textbox, when the data you are displaying is not applicable in for the current state of a dialog or window."

However, I will change it if you insist.

Mike E

-- 
PGP key:
http://pgp.mit.edu:11371/pks/lookup?op=get&search=0x00CDB13500D7AB53

rocketman768:
Thanks for thinking about UI design principles.

I don't think we have a use case for any copy/pasting of any of the text fields. Also, is the value of the preboil volume relevant in the editor when the checkbox is ticked? I think I would be just as happy if the textbox disappeared when the checkbox is ticked, indicating to me that disabling it is the best option.

src/EquipmentEditor.cpp
EvansMike: OK. Pushed.

EvansMike:
OK. Pushed.

rocketman768:
:+1:

PR #173, opened 2016-02-08
Fixes #159: OG in Plato wrongly displayed in refracto dialog

### LINKED ISSUE #159: [stable/2.3.0] OG °P wrongly calculated for OG sg in refractometer tool
If you enter the OG in sg then you enter the current gravity in °P, after clicking on calculate, the OG value in °P is fill with a wrong value (444 851,457 P for an OG of 1.067 sg in my case).

This is a minor bug as it doesn't prevent the tool to compute all this other values correctly.

rocketman768:
@mikfire did I do this right?


rocketman768:
mik says :+1: in #172

PR #176, opened 2016-02-08
Closes #175: Save/Restore UI state

### LINKED ISSUE #175: Save/Restore UI state
I want to preserve changes to UI layout made by the user. 
I'll save/restore the state of MainWindow splitters, and treeView/tableView headers.

danielm5:


rocketman768:
Sweet! Are we also saving the overall geometry of the main window? If not, could we (by using [saveGeometry/restoreGeometry](http://doc.qt.io/qt-5/restoring-geometry.html))?

Once that is done, go ahead and squash all these commits together and I'll merge.

theophae:
That's really a good thing ! Thank you Daniel.

danielm5:
MainWindow geometry and state was already being saved.

rocketman768:
:+1:

PR #178, opened 2016-02-10
Fixes #177: first wort hop adjustment 100x too high

### LINKED ISSUE #177: First wort hop adjustment too high by a factor of 100
For example, half an ounce of citra in a 5.5 gallon batch with a 60 min boil outputs a reasonable 19.5 IBU, changing that to first wort results in 1,948.2 IBU, which seems high.

image
image

Edit: I just found the IBU adjustments and they're set to 10,000 for FW and mash. I have not fiddled with these, but I did change units from metric to imperial at the start. Seems like this is just a bad initial config. Also, the units for this adjustment are unclear that it's percent and not a multiplier.

rocketman768:


mikfire:
The last time I researched this, the literature suggests the default should be 1.1. Which is really, really picking nits :)

rocketman768:
ermagerd mik! Nah, I should do it properly...in recipe.cpp it defaults to 1.1 so I'll make it consistent with that.

PR #181, opened 2016-02-11
Fixes #180: crash if moveUP or moveDown in empty NameMashEditor dialog

### LINKED ISSUE #180: Program crashes after moving up or down a mash step for an empty mash profile
Actions:
1 - Open Brewtarget
2 - Click on "Mashes" at the toolbar
3 - Remove the currently selected Mash
4 - Notice that no data for any mash will be visible
5 - Click on the "Mash step up" or "Mash step down"

Result:
Program crashes on Windows.

crash

danielm5:


danielm5:
We should check if similar bugs affects other dialogs.

rodrigoA:
Yes, definitely. Actually, IMHO we should also disable buttons when there are no items available for such actions.

rocketman768:
Good catch @danielm5. Merging this now to avoid the crash. The UI cleanup as mentioned by @rodrigoA would be a good enhancement if anybody wants to take it on.

rodrigoA:
Yeah, I can do that. As soon as I get my enviroment set up I will do that.

Issue #179, opened 2016-02-10
2.3.0 usability


rocketman768:
Below is an email I received. Putting it here for visibility.

I thought i'd send you some feedback and bugs etc…

Clicking OK or CANCEL from a HOP EDITOR window causes the HOP DATABASE window to move behind the main GUI window (very annoying). Seems to happen with the FERMENTABLE DATABASE too.

Entering weight amounts into the HOPS section now defaults to kg instead of g. I can't figure out a way of changing that.  In v2.1 i could click on the AMOUNT collumn of the hop i wanted to change and just type "25" for example, and hit enter, it would register 25 grams. Now i need to enter "25g", which is not a huge thing, but a tad annoying as a home brewer… 25kg of Horizon is probably hard to come by also :P

Also entering times in the hops tab is now a pain in the ass… if i enter 1.5 hr on my first wort hops (with my boil time set for 90 mins), then hit enter, the value jumps to 2 hr. And selecting SCALE > minutes is no good, as then my dry hop additions are 5,760 min instead of 4 days for example.

The former "2.8 L" reading of my Joe White Ale Malt for example, in the COLOUR column of the FERMENTABLES tab is now "3 srm"… everything is rounded up to the next whole number, is there a way to see the more precise readout in Lovibond instead of the srm?

Is there a way to save/add a setup in the EQUIPMENT EDITOR instead of having to type it all out each time? I mean boil/batch sizes, water usage, mash tun size etc… name the profile "Paintbucket" for example, then save?

i feel the SUBSTITUTES box in the HOP EDITOR needs to be smaller and the NOTES box needs to be bigger. Most of my hops have one or two substitutions, but a fair bit of information in the notes selection.

Thanks again for your time, awesome stuff!!

theophae:
Regarding the problem of the default unit (like the kg for hops), it could be a good idea to add a page in the options for the user to choose these favourites units.
About the problem of time units, I had the same feeling about the default time display. To me it is highly unusable to deal with decimals of hours, and the time for dry hop shows another problem with this system.
To solve this, we could add to the future unit tab in the options different choices like "time unit for hop boil addition", "time unit for hop aroma addition", "time unit for dry hop", "time unit for mash steps", and so on.
This tab for unit config would make every on happy

mikfire:
"[D]o not forget that you SHOULD enter the unit suffix (lb, kg, L, mL,
etc.) for the amount boxes in brewtarget. If you do not, brewtarget will
[...]

PR #186, opened 2016-02-12
Scale on input -- item #179

mikfire:
Assuming I did not utterly and completely make a mess of it, this should allow
us to scale input. What I mean by that is if you specify the scale of a field
to be (for example) grams, and then type 19 into that field, brewtarget will
now interpret this as 19 grams, not 19 kg.

The unit&scale mechanisms were about 90% in place. I had to modify a few
method signatures, a few method invocations in the tables and ONE LINE in
BtLineEdit. Everything should play by the rules now. 

I should also mention I had to add the necessary code to the WaterTableModel to deal with units and scales. One day, I really want to take the time to make that stuff work.


rocketman768:
@theophae @EvansMike 

want to build and try it before I merge?

theophae:
I'll have a look to it. I hope today

theophae:
I've just tested it and it is perfect :)
Really cool :+1:

PR #191, opened 2016-02-14
Fix #183 amountIsWeight behavior in YeastEditor

### LINKED ISSUE #183: Display amount always in volumeunit in yeast editor
Display amount is always in volume unit whereas AmountIsWeight checkbox is selected in yeast editor.

theophae:
Is it OK like this ? I rebased my branch to be based on stable/2.3.1.


rocketman768:
:+1: perfect

PR #192, opened 2016-02-15
Closes #117: New ingredient can't be created into a folder

### LINKED ISSUE #117: New ingredient can't be created into a folder
If you create a folder for any type of ingredient (fermentable, hop, misc...) and you right click on this folder and choose to create a new ingredient, it won't be created into the folder, but it will appear at the end of the list. The only way to add it to the folder is to drag and drop it on the folder (which is really a pain)

mikfire:
It wasn't so much a bug as an intentional design choice. The problem I avoided
solving was that the tree didn't know which editor to call.

I fixed that by modifying the signal to call a BtTreeView slot, which figured
out the folder and called the proper editor. It required me adding a new
method to all of the editors, and modifying the signal slightly.

### DIFF
--- a/src/BtTreeView.cpp
+++ b/src/BtTreeView.cpp
@@ -39,6 +39,12 @@
 #include "yeast.h"
 #include "brewnote.h"
 #include "style.h"
+#include "FermentableDialog.h"
+#include "EquipmentEditor.h"
+#include "HopDialog.h"
+#include "MiscDialog.h"
+#include "StyleEditor.h"
+#include "YeastDialog.h"
 
 BtTreeView::BtTreeView(QWidget *parent, BtTreeModel::TypeMasks type) :
    QTreeView(parent)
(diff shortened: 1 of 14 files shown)


rocketman768:
@theophae does this fix your #117 ?

theophae:
Yes, it fixes it :+1:

src/BtTreeView.cpp
```diff
@@ -351,19 +393,19 @@ void BtTreeView::setupContextMenu(QWidget* top, QWidget* editor)
...
-         _newMenu->addAction(tr("Hop"), editor, SLOT(newHop()));
+         _newMenu->addAction(tr("Hop"), this, SLOT(newIngredient()));
          break;
       case BtTreeModel::MISCMASK:
-         _newMenu->addAction(tr("Misc"), editor, SLOT(newMisc()));
+         _newMenu->addAction(tr("Misc"), this, SLOT(newIngredient()));
          break;
       case BtTreeModel::STYLEMASK:
-         _newMenu->addAction(tr("Style"), editor, SLOT(newStyle()));
+         _newMenu->addAction(tr("Style"), editor, SLOT(newIngredient()));
```
theophae: I've seen a warning about this line:
`QObject::connect: No such slot StyleEditor::newIngredient() in /home/theo/brewtarget_dev/src/BtTreeView.cpp:408
QObject::connect:  (receiver name: 'styleEditor')`

It seems to me that it should be 
`C++
_newMenu->addAction(tr("Style"), editor, SLOT(newStyle()));`

instead of:
`C++
_newMenu->addAction(tr("Style"), editor, SLOT(newIngredient()));`

mikfire: Hmm. No that's just a mistake. The slot should be newIngredient() but it
should be binding to the tree, not the editor. I'll fix.

Mik

mikfire:
Odd. That's not showing in my branch?

mikfire:
Ok. Figured out what I wasn't seeing. Previous branch wouldn't have fixed yeast or style. 00e8b68 should.

rocketman768:
Good now @theophae ?

theophae:
Yes :+1:

Issue #88, opened 2015-09-27
Printed pages do not fill page width on high-dpi displays


rocketman768:
I am using higher font DPI (150) since I have a large resolution.

![printed_recipe](https://cloud.githubusercontent.com/assets/752558/10123758/ba921ccc-64f6-11e5-9b2e-db2593751095.png)

mikfire:
I don't have a hi-res display , but I suspect this may be because css/recipe.css defines the width as 8.5in. I do all the rest as proportional widths. I did that, as I recall, because somebody wanted margins when they printed it.

Just for grins ( I cannot test this myself ), try increasing that to 10.5in, recompiling and see what happens.

rocketman768:
You can't change your DPI? I know this is possible on Windows and KDE.

rocketman768:
Something like

    body{
       width: 90%;
       margin: auto auto auto auto;
    }

seems to be somewhat close. The font sizes still need to be specified in em or pt or whatever doesn't get smaller with higher dpi.

PR #194, opened 2016-02-17
Addresses issue #88

mikfire:
Simply tweaking the CSS to use EM instead of px. I really can't test this, so
somebody with a highres display needs to test and make sure.


rocketman768:
Closing in favor of #278.

PR #197, opened 2016-02-20
Closes #195: half of recipes moved on folder drag-n-drop

### LINKED ISSUE #195: Half the recipes in a folder are moved when dragged-and-dropped into another folder
I am experiencing some strange behavior in version 2.3.0.

When a folder containing recipes is dragged inside another folder only some of the recipes in the folder are moved. 

At first it appears as though the recipes not moved have disappeared, but after saving and closing Brewtarget the recipes will be in there original location when the software is reopened.

I attached a couple screen shots showing what I am talking about.

recipes

mikfire:
Moving folders used childCount() as the test condition in a for loop. Because
I moved the children into the new folder, childCount() got decremented by one.
This resulted in the loop counter getting incremented, the test condition
getting decremented and me moving only about half the children.

The fix is a bit more involved, since I don't move folders. I capture the
starting childCount at the start and use that for the test condition, and I
only take the first non-folder child through the whole operation. The only
tricky part is that I don't move folders, so I have to increment a counter to
indicate where the first non-folder item is.

### DIFF
--- a/src/BtTreeModel.cpp
+++ b/src/BtTreeModel.cpp
@@ -815,7 +815,7 @@ void BtTreeModel::copySelected(QList< QPair<QModelIndex, QString> > toBeCopied)
             
             break;
          default:
-            Brewtarget::logW(QString("copySelected:: unknown type %1").arg(type(ndx)));
+            Brewtarget::logW(QString("deleteSelected:: unknown type %1").arg(type(ndx)));
       }
    }
 }
@@ -829,6 +829,9 @@ void BtTreeModel::deleteSelected(QModelIndexList victims)
       QModelIndex ndx = toBeDeleted.takeFirst();
       switch ( type(ndx) ) 
       {
+         case BtTreeItem::RECIPE:
+            Database::instance().remove( recipe(ndx) );
+            break;
          case BtTreeItem::EQUIPMENT:
             Database::instance().remove( equipment(ndx) );
             break;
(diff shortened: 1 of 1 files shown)


src/BtTreeModel.cpp
```diff
@@ -841,12 +844,6 @@ void BtTreeModel::deleteSelected(QModelIndexList victims)
          case BtTreeItem::MISC:
             Database::instance().remove( misc(ndx) );
             break;
-         case BtTreeItem::RECIPE:
-            Database::instance().remove( recipe(ndx) );
-            break;
-         case BtTreeItem::STYLE:
-            Database::instance().remove( style(ndx) );
-            break;
```
rocketman768: Styles are no longer handled in `deleteSelected()`?

rocketman768:
Also, please rebase and repush to stable/2.3.1

mikfire:
Sorry, git and I don't get along.

I'm happy to rebase but what is the sequence of git commands?

Mik

src/BtTreeModel.cpp
mikfire: I have exactly 0 clues what happened there. As soon as I figure out how to "rebase and repush", I will fix this.

rocketman768:
$ git fetch upstream
    $ git checkout bugs/foldermove
    $ git rebase upstream/stable/2.3.1
    $ git push --force origin bugs/foldermove

This will rebase the commit on top of upstream/stable/2.3.1. You will probably need to close this PR and open a new one against Brewtarget:stable/2.3.1 instead of Brewtarget:develop.

rocketman768:
For future bugs that pertain to stable/2.3.1, you can base the fix directly on stable/2.3.1 at the start:

    $ git branch bug/something-bad upstream/stable/2.3.1
    $ git checkoug bug/something-bad

PR #199, opened 2016-02-21
Closes #195 again

### LINKED ISSUE #195: Half the recipes in a folder are moved when dragged-and-dropped into another folder

mikfire:
The rebase generated way too many collisions. It was faster to create a new
branch.

Moving folders used childCount() as the test condition in a for loop. Because
I moved the children into the new folder, childCount() got decremented by one.
This resulted in the loop counter getting incremented, the test condition
getting decremented and me moving only about half the children.

The fix is a bit more involved, since I don't move folders. I capture the
starting childCount at the start and use that for the test condition, and I
only take the first non-folder child through the whole operation. The only
tricky part is that I don't move folders, so I have to increment a counter to
indicate where the first non-folder item is.

### DIFF
--- a/src/BtTreeModel.cpp
+++ b/src/BtTreeModel.cpp
@@ -989,7 +989,7 @@ bool BtTreeModel::renameFolder(BtFolder* victim, QString newName)
    QPair<QString,BtTreeItem*> f;
    QList<QPair<QString, BtTreeItem*> > folders;
    // This space is important       ^
-   int i;
+   int i, kids,src;
 
    if ( ! ndx.isValid() )
       return false;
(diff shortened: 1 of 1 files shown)


mikfire:
This is against stable/2.3.1. As the checkin says, I tried the rebase but the number of collisions was huge, and every time I attempt conflict resolution with git I make a huge mess. So I created a new branch, closed the old pull request and created this one

rocketman768:
Sometimes cherry-picking can work better.

This PR still says it's against develop.

mikfire:
Cause I'm an idiot. Let me see if I can fix this.

PR #202, opened 2016-02-29
Split install into data and runtime components

rocketman768:
Change is necessary to enable distinct data and binary packages.
Specifically, this is to avoid the lintian warning
[arch-dep-package-has-big-usr-share](https://lintian.debian.org/tags/arch-dep-package-has-big-usr-share.html)


rocketman768:
Hearing no objections, :+1: myself

Issue #63, opened 2015-07-28
Final Batch Sparge always assuming 15min in Mash Wizard


renato:
I created a mash configuration that has a "Final Batch Sparge" of 20 min. It is not a big deal, but even if I have a "Final Batch Sparce" with this duration, the wizard always assumes 15 min:
https://github.com/Brewtarget/brewtarget/blob/master/src/MashWizard.cpp#L291

Another point that is not a big deal but sounded strange to me is that I changed the name "Final Batch Sparge" to "Fly sparge" (which is what I do) and the Mash Wizard ended up creating the step again:
https://github.com/Brewtarget/brewtarget/blob/master/src/MashWizard.cpp#L251

I think that it'd be interesting to set a mash step as the sparge step (a new checkbox or new infusion type) so the Mash Wizard wouldn't need to compare the step name with a hardcoded string.

Plus, if a mash step for sparge is already present, its time should be used instead.

theophae:
I also think that to add a checkbox to indicate the sparge step could be a good idea.

dpettersson:
I feel that a little more information is needed in order to be able to implement this.

The hardcoded value of 15 minutes is a simple thing to change. The name is also simple, but it raises a question if we need to be able to handle more languages (the name of the sparge is hardcoded and thus only suports English). The idea about the checkbox could solve the name problem, but can the function be used for anything other then batch sparging?  It looks like the wizard recalculates (all?) the needed volumes when used.

dkoppenh:
Also, please consider the case for no-sparge / BIAB.  A checkbox y/n would help with that.

dpettersson:
It feels like there is a need to extend the wizard with som options.

rocketman768:
Yeah, there needs to be a total rework of the mash wiz into a real wizard dialog with options. Sparge, no-sparge, BIAB, internationalization, etc. need to be included.

theophae:
Wouldn't it be a good idea to add "Sparge" to the mash type in addition to "Infustion" "Temperature" and "Decoction"  ?
This won't replace all the work on the mash wizard, it is just another option than adding a checkBox.

mikfire:
If you want, I have a branch called feature/newmashtype with 90% of the
work done. It has the code for two new sparge types: flySparge and
batchSparge. There is, as far as I can tell, just one outstanding bug that
I was never able to squash.

The issue that I cannot resolve is when using the mash designer. If you
create (for example) two sparge steps at the end and then close the mash
[...]

PR #205, opened 2016-03-03
Enhancement -- Adding two new mash types

mikfire:
This is an enhancement I've been working on for a while, but I think I finally got it right. I've added two new mash types: batchSparge and flySparge. The are considered by most of the code as infusion steps (which they basically are), but they do behave a little differently -- batchSparge assumes you've emptied the mash tun  for example, and adding a flySparge step does not allow you to continue.

This should also close the issue #63


rocketman768:
There are some conflicts to resolve

    $ git fetch upstream
    $ git merge upstream/develop
    ...carefully fix conflicts...
    $ git push origin feature/newmashtypev2

mikfire:
Did that work? For reasons that escape me, I do not grok git conflict resolution.

rocketman768:
The merge commit looks good. Seems that the conflicts were some pretty trivial stuff in database.c/cpp?

mikfire:
Yeah. The postgresql changes moved some bits around, and this change added
a new parameter in the same places in the header. The conflict in the cpp
file was even more trivial.

I thought I had synced my master branch to the development branch, but
apparently hadn't.

Mik

PR #207, opened 2016-03-03
Closes #206 - core while populating child tables

### LINKED ISSUE #206: Core dump while populating child tables
If you start with a clean install, the most recent HEAD code will dump core. I've a fix I will commit in just a few minutes

mikfire:
populateChildrenTables() requires the tableName hash to be populated, but
the hash can't be populated until after DatabaseSchemaHelper::migrate is
called. With the way the methods are written now, this leads to a chicken-egg
sort of problem.

I solved the problem by simply have populateChildrenTables() prepopulate the
hash. On the one time we need to call that method, the hash will get
initialized twice. I could maybe fix that, but it seemed an awful lot of work
for something that should happen once. In a broader question, shouldn't the
populateChildrenTables be part of the migrate code?

The changes to the header files are simply cleaning some compiler warnings.

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -310,14 +310,16 @@ bool Database::load()
          }
    }
 
-   // With the db open, we can get all our hashes. This has to happen before
-   // updateSchema
-   Database::tableNames = tableNamesHash();
-   Database::classNameToTable = classNameToTableHash();
    // Update the database if need be. This has to happen before we do anything
    // else or we dump core
    bool schemaErr = false;
    schemaUpdated = updateSchema(&schemaErr);
+
+   // Since updateSchema could add new tables, we have to wait until this
+   // point to populate the tables. 
+   Database::tableNames = tableNamesHash();
+   Database::classNameToTable = classNameToTableHash();
+
    if( schemaErr )
    {
          QMessageBox::critical(
@@ -327,7 +329,6 @@ bool Database::load()
          );
          return false;
    }
-
 
    // Initialize the SELECT * query hashes.
    selectAll = Database::selectAllHash();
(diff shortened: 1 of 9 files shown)


src/database.cpp
```diff
@@ -310,14 +310,16 @@ bool Database::load()
          }
    }
 
-   // With the db open, we can get all our hashes. This has to happen before
-   // updateSchema
-   Database::tableNames = tableNamesHash();
-   Database::classNameToTable = classNameToTableHash();
```
rocketman768: Is the comment

> This has to happen before updateSchema

just wrong?

mikfire: It is now. I will fix.

Mik

mikfire: It is mostly wrong, but being mostly wrong is like being mostly dead.

We have to populate the hashes before we call populateChildrenTables(), which is called by updateSchema(). The problem was that until DatabaseSchemaHelper::migrate() is called to do the v5 -> v6 upgrade the table that populates the hash doesn't exist. This causes populateChildrenTables to throw an exception.

I fixed this the first time by changing the order and populating the hashes first. That is when I wrote the original comment. It was right at the time, but only in the very limited circumstances of my test systems.

When I realized the true nature of the problem, it also occurred to me that we have to wait until after updateSchema() is called, since it may add tables to the database. Populating the hashes before wouldn't see the new tables.

mikfire:
PS -- I am not 100% sure using the table in the database to populate the hash was one of my good ideas. It has some significant advantages, but it also has some significant disadvantages.

rocketman768:
Well it's weird, but it is now documented. Let's keep thinking about how to untangle all the order dependencies we have at startup in the db.

PR #215, opened 2016-03-18
Closes #214 -- error in equipment editor definitions

### LINKED ISSUE #214: Error in equipment editor definitions
About line 251 in EquipmentEditor.cpp, I don't think the editField is supposed to be topUpKettle_l. I think it should be "topUpWater_l". This means editing the Final Top-up field will set the Kettle Top-up field.

mikfire:
Simple fix -- make certain the edit field refers to the actual field being edited. This is kind of a hot bug.

Any chance I can talk anybody into reverting to the UI files until I can get the QML files working?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,7 +126,7 @@ ENDIF()
 #==================================Find Qt5====================================
 
 # The minimum QT5 version we need. Keep as low as necessary.
-SET(QT5_MIN_VERSION 5.3)
+SET(QT5_MIN_VERSION 5.1)
 
 # Automatically run moc on source files when necessary
 SET(CMAKE_AUTOMOC ON)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -126,7 +126,7 @@ ENDIF()
 #==================================Find Qt5====================================
 
 # The minimum QT5 version we need. Keep as low as necessary.
-SET(QT5_MIN_VERSION 5.3)
+SET(QT5_MIN_VERSION 5.1)
```
rocketman768: This is fine, but put it in another PR.

mikfire: That was an accident. Didn't mean to commit it. Anyway you can just ignore that one?

rocketman768: Please rework, and I'll submit.

PR #216, opened 2016-03-19
Use vagrant for development

speshak:
I added Vagrant infrastructure to make it easier to try out the new Postgres support.  This PR includes an Ansible playbook which configures the machine with the correct version of Postgres & configures the database.

Documentation is included in dev-doc/vagrant.markdown

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ build/*
 # Kate project
 .kateproject.d/
 .idea
+
+.vagrant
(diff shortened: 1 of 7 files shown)


mikfire:
Oh. That's brilliant. I hadn't even thought to include something like that.

speshak:
I'm not sure how any of my changes could have broken the build.  If anyone with more knowledge of the project has any insight that would be great.

rocketman768:
It's not you. It's #217

PR #224, opened 2016-03-22
Closes #223 Changing IBU formula or mash/fwh percentages, nothing happens

### LINKED ISSUE #223: Changing IBU formula or mash/fwh percentages, nothing happens
To get things to update, you need to close brewtarget and reopen it. As far as I can tell, it's always been this way. So I'm not sure if it is a feature request or a bug. Anyway, I've a commit for it.

mikfire:
Right now, if you modify the hop usage percentages or the IBU formula, you
have to restart brewtarget. This was surprisingly annoying to me, so I fixed
it.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -884,10 +884,16 @@ void MainWindow::showChanges(QMetaProperty* prop)
 
    // See if we need to change the mash in the table.
    if( (updateAll && recipeObs->mash()) ||
-       (propName == "mash" &&
-       recipeObs->mash()) )
+       (propName == "mash" && recipeObs->mash()) )
    {
       mashStepTableModel->setMash(recipeObs->mash());
+   }
+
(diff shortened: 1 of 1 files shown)


src/MainWindow.cpp
```diff
@@ -884,11 +884,17 @@ void MainWindow::showChanges(QMetaProperty* prop)
...
+       (propName == "mash" && recipeObs->mash()) )
    {
       mashStepTableModel->setMash(recipeObs->mash());
    }
+
+   // Not sure about this, but I am annoyed that modifying the hop usage
+   // modifiers isn't automatically updating my display
+   if ( updateAll ) {
+     recipeObs->acceptHopChange( recipeObs->metaProperty("hops"), QVariant());
+     hopTableProxy->invalidate();
```
rocketman768: Very strange. We don't have an actual mechanism to propagate this info?

PR #226, opened 2016-03-22
Make running instance check less Linux specific

speshak:
Someday I'd like to get Brewtarget to build on BSD.  This is one very tiny step towards that goal.

This does change the behavior slightly in that the single-instance restriction is only enforced if the user is using SQLite.  I don't think there is any harm in allowing multiple instances to run against Postgres. (Although I may doubt the usefulness)


rocketman768:
:+1:

PR #196, opened 2016-02-18
Run the tests as part of the travisCI script

rocketman768:

PR #229, opened 2016-03-26
Fixes #210: Missing dirs fatal to initialization

### LINKED ISSUE #210: [develop] Brewtarget::initialize() fails on a blank system
With no .config directory or /usr/share/brewtarget, Brewtarget poops itself and gives up at initialization.

speshak:
Refactor of Brewtarget::ensureDirectoriesExist and related methods so
that:
1) Required directories are created at startup
2) Failure to create said directories is no longer fatal to
   Brewtarget::initialize


speshak:
I also have a rebase of #196 working with some additional changes.

Issue #237, opened 2016-04-19
Deleting equipment, style or named mash dumps core


mikfire:
This is an oversight from my database work. I somehow didn't configure some required properties on these three items. I have a fix ready.

mikfire:
Closed with merged commit ae27159

PR #240, opened 2016-04-19
Closes issue #237 -- Deleting equipment dumps core

mikfire:
In all the work I did on the database, I somehow forgot to add the two
required properties to these three items. This patch fixes that oversight.


rocketman768:
:+1:

Issue #239, opened 2016-04-19
fromXML methods are not fault tolerant


mikfire:
Many other projects are not implementing the beerXML specification properly. In several places, we expect to map directly from a text string to a data type without testing first. When a faulty beerXML export uses "Base Malt" as a fermentable type, we can't map it and promptly dump core.

In every case where we are dependent on mapping XML input to a type, we need to test the map before we simply rely on it working. And yes, I have a fix.

mikfire:
merged commit ebcf76d

PR #241, opened 2016-04-19
Closes issue #239 -- fault tolerant fromXML

mikfire:
I simply make sure we can map from any string to a type. This impacts a few
places, but the code it pretty simple.


rocketman768:
:+1:

PR #242, opened 2016-04-20
Closes #238 -- Units on input are not working

### LINKED ISSUE #238: Units on input are not working
Regardless of what unit you specify on input, it is always "interpreted" as the system default. This is a side effect of the input scaling work I did. I have a fix, but I would like to make sure I understand what side effects my solution may have.

mikfire:
For reasons I don't understand (other than I suck at testing code), I never
tested a BtLineEdit to see how the new input scaling works. In short, it
didn't. No matter what you entered, the default unit/scale was displayed. So
if you entered "20 qt" into a BtVolumeEdit, it would be interpreted as 20 L or
20 gallon, depending on what the field was set to.

The fix ends up being rather simple. BtLineEdit::toSI() invokes the proper logic 
to find the default unit/scale if it is called with force = false. I just had to figure out when
to invoke that logic in BtLineEdit::lineChanged().

After much screwing around, I figured out that we needed to invoke it only when
the user didn't include a unit/scale in the input. So if the user says "20 qt"
in a BtVolumeEdit, we won't go looking for the default unit/scale. If they just
say "20", we will.

I will be happy if I don't have to mess around with BtLineEdit for a while.

### DIFF
--- a/src/BtLineEdit.cpp
+++ b/src/BtLineEdit.cpp
@@ -58,7 +58,7 @@ void BtLineEdit::lineChanged(Unit::unitDisplay oldUnit, Unit::unitScale oldScale
    // This is where it gets hard
    double val = -1.0;
    QString amt;
-   bool force = false;
+   bool force = true;
    bool ok = false;
 
    // editingFinished happens on focus being lost, regardless of anything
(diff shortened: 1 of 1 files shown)


rocketman768:
:+1:

PR #245, opened 2016-04-23
Closes #244 -- New mash type are behaving poorly

### LINKED ISSUE #244: New mash type are behaving poorly
I just created a new recipe. If I do a double batch sparge with a grist ratio of 1.5 qt/lb, I get a warning saying that there is more liquid than mash tun. If I do a double batch sparge at a grist ratio of 2, no warning and the math works.

The sort order is also a little out of whack. 

Finally, I missed a few qDebug statements in there somewhere.

mikfire:
I merged master and couldn't reproduce the problem I was seeing. I did find
another problem with the no sparge math, so I fixed that.

Then I decided the new mash wizard screen looked like crap, so I fixed that. I
rather like the new form better. It almost makes some kind of sense.

And then I decided I really wanted the wizard to have a lot more magic. So now
it figures out what the last kind of sparge was, how many of them there were
and what the original grist ratio and sets all the fields appropriately.

I even remembered to remove the qDebug() statements.

### DIFF
--- a/src/MashWizard.cpp
+++ b/src/MashWizard.cpp
@@ -44,7 +44,7 @@ MashWizard::MashWizard(QWidget* parent) : QDialog(parent)
    bGroup->addButton(radioButton_batchSparge);
    bGroup->addButton(radioButton_flySparge);
 
-   radioButton_batchSparge->setChecked(true);
+//   radioButton_batchSparge->setChecked(true);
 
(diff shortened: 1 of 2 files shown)


mikfire:
The problem I am trying to fix seems to be dependent on using postgresql, not sqlite. So the problem wasn't fixed, just hidden. I am closing this pull request until i fix it

PR #246, opened 2016-04-25
Closes #244 -- New mash type are behaving poorly (for reals)

### LINKED ISSUE #244: New mash type are behaving poorly

mikfire:
According to both sqlite and postgres, the order in which rows are returned is undefined unless an order by clause is used. In postgresql, it seems "undefined" means random. In sqlite, it seems "undefined" means in row-id order.

When we got the mash steps, we didn't specify the order. It happened to work for sqlite, but not for postgresql. This patch adds an order by clause to make sure the order isn't "undefined".

I've also really horked the mash wizard interface around. Having brewed with it, I like it.

### DIFF
--- a/src/MashWizard.cpp
+++ b/src/MashWizard.cpp
@@ -44,7 +44,7 @@ MashWizard::MashWizard(QWidget* parent) : QDialog(parent)
    bGroup->addButton(radioButton_batchSparge);
    bGroup->addButton(radioButton_flySparge);
 
-   radioButton_batchSparge->setChecked(true);
+//   radioButton_batchSparge->setChecked(true);
 
    connect(bGroup, SIGNAL(buttonClicked(QAbstractButton*)), this, SLOT(toggleSpinBox(QAbstractButton*)));
    connect(buttonBox, SIGNAL(accepted()), this, SLOT(wizardry()) );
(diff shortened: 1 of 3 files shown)


rocketman768:
Can you post a screenshot and point out the changes?

mikfire:
Sure. 


 old dialog 
The previous incarnation was basically an expansion on the original. I didn't like it. There were entry fields, then some buttons and then another entry field. If we selected no-sparge, did we have to still put in a grist ratio? It was scattered and not behaving like a proper work flow. The dialog window itself was just too large.

new batch sparge dialog 
The new layout is more compact. Buttons are grouped with buttons, and inputs are grouped with inputs. Since both inputs were numeric only, it made sense to make them spin boxes. It works like a proper work flow -- first tell me what kind of sparge, then give me the information I need for that sparge type. It is left-to-right; my apologies to those languages that read right-to-left.

If you select fly sparge, the batches spinBox is disabled.
look ma, no batches  

If you select no-sparge, both spin boxes are disabled.
both spinboxes disabled 

If you open the mash wizard on a recipe with an existing mash, the dialog does two things in this patch. First, it recalculates the previous grist ratio and uses that to prepopulate the grist ratio spin box. Second, if you are batch sparging it counts the number of batch sparges and uses that to initialize the batches spinbox. I cannot tell you how long I've wanted that feature.

rocketman768:
:+1:

PR #249, opened 2016-05-01
Bug #1109730 - Hydrometer Tool

rhoob:
Add a Hydrometer adjustment tool on the tool menu.  Adapted from the ConverterTool.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -98,6 +98,7 @@ SET( brewtarget_SRCS
     ${SRCDIR}/HopEditor.cpp
     ${SRCDIR}/HopSortFilterProxyModel.cpp
     ${SRCDIR}/HopTableModel.cpp
+    ${SRCDIR}/HydrometerTool.cpp
     ${SRCDIR}/instruction.cpp
     ${SRCDIR}/IbuGuSlider.cpp
     ${SRCDIR}/IbuMethods.cpp
@@ -237,6 +238,7 @@ SET( brewtarget_MOC_HEADERS
     ${SRCDIR}/HopEditor.h
     ${SRCDIR}/HopSortFilterProxyModel.h
     ${SRCDIR}/HopTableModel.h
+    ${SRCDIR}/HydrometerTool.h
     ${SRCDIR}/IbuGuSlider.h
     ${SRCDIR}/InstructionWidget.h
     ${SRCDIR}/Log.h
--- a/src/HydrometerTool.cpp
+++ b/src/HydrometerTool.cpp
@@ -115,19 +115,20 @@ void HydrometerTool::convert()
    /*
    cg = corrected gravity
    mg = measured gravity
-   tr = temperature at time of reading
-   tc = calibration temperature of hydrometer*/
+   tr = temperature at time of reading in Fahrenheit
+   tc = calibration temperature of hydrometer in Fahrenheit*/
 
    double cg;
    double mg;
-   QString tr_string;
-   QString fahr = "F";
    double tr;
    double tc = 60;
    bool ok = false;
 
-   tr_string = Unit::convert(lineEdit_inputTemp->text(), fahr);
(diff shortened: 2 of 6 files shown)


mikfire:
I am a little new to these code reviews. If I seem to be being rude, it is completely unintentional. 

In the copyright, please list yourself as the author and grant copyright to rocketman. You can use BtLineEdit as an example of how I've done it in the past. 

inputLineEdit should be a BtDensityEdit with the forcedUnit property set to "displaySg". This will get a lot of the magic formatting right, without causing problems. You can do this with a line like:

    inputLineEdit->setProperty("forcedUnit",QVariant(QStringLiteral("displaySG")));

I would recommend doing the same for outputLineEdit. You just need to set the field to read-only. You can look at lineEdit_boilSg for an idea of how to make it work. This will also allow you to simply use outputLineEdit->setText() and have all the conversions done.

I would be happier if you renamed the fields to follow our loose naming convention (lineEdit_[description]) but I am not really going to insist on it.

inputTempLabel should be a BtTemperatureLabel, so the user can select how they measure temperature with a right click. This will require you to set connect the labelChanged signal. EquipmentEditor.cpp around line 97 is a good example for how to do that.

This means you will also need to add the configSection property on the formLayout. You can see hints on how to do this in EquipmentEditor.cpp around line 140

You should also use inputTempLineEdit->toSI() to get the temperature the user entered, automatically converted to C. I will be honest and admit I have no idea how this will mess up your formula.

Finally, if you got your final formula from a source, could you please add a reference as a comment just prior to the maths? Again, I won't insist but it will help us later when somebody disputes a result.

rhoob:
Thanks for the feedback.  I'm new to everything.  My background is Electrical Engineering and I'm working on my programming knowledge, so the feedback is great learning for me.

I've made all of the suggested changes, but during my testing the BtTemperatureLabel does not appear to work correctly.  I was looking at the label_boilingPoint in the EquipmentEditor.cpp.  and it also does not appear to work.  When I right click and change between Celsius and Fahrenheit, nothing in the lineEdit_boilingPoint changes.  I expected that the lineEdit_boilingPoint would update to the selected unit.

mikfire:
I think this is a known problem that I fixed, but I can't find the merge
request at the moment. A signal wasn't connected that should have been.
Since I can't find the merge request, let me make sure of this.

Mik

On Sun, May 15, 2016 at 5:32 PM, rhoob <notifications@github.com> wrote:

> Thanks for the feedback. I'm new to everything. My background is
> Electrical Engineering and I'm working on my programming knowledge, so the
> feedback is great learning for me.
>
> I've made all of the suggested changes, but during my testing the
> BtTemperatureLabel does not appear to work correctly. I was looking at the
> label_boilingPoint in the EquipmentEditor.cpp. and it also does not appear
> to work. When I right click and change between Celsius and Fahrenheit,
> nothing in the lineEdit_boilingPoint changes. I expected that the
> lineEdit_boilingPoint would update to the selected unit.
>
> —
> You are receiving this because you commented.
> <https://github.com/Brewtarget/brewtarget/pull/249#issuecomment-219311882>
>



-- 

mikfire:
@rhoob, what you need to do on that is to create the inputTempLabel as a BtTemperatureLabel, and then setContextMenuPolicy(Qt::CustomContext).

The EquipmentEditor.cpp is the best example for doing this. 

if you want to get the right click menu working, I will hold off on merging this request. If you are good with it, I will merge. My apologies for taking so long to finish the review.

rhoob:
No problem on the time.  Thanks for looking it over.

I think this is what I have done.  label_inputTemp is a BtTemperature as declared on line 49 and 50.  (It is declared twice.  I'll fix that.)  Then on line 52 is setContextMenuPolicy(Qt::CustomContextMenu).  I think you meant CustomContextMenu instead of CustomContex.  The right click menu does not seem to work on the equipment editor page either.

PR #254, opened 2016-05-16
WIP: do not dirty database on recipe load.

greggree:
I'm pretty sure we need to calculate the FG and OG. When we load a recipe up in the app, it always calls recalcAll() with causes the FGOG calc to set the og and fg. This causes the Database::makeDirty() to get called. Now when you start the app, and then close it, it doesn't request a database save message.


src/recipe.cpp
```diff
@@ -1839,15 +1839,22 @@ void Recipe::recalcOgFg()
       // being updated for no good reason.
       // NOTE: We are we recalculating all of these on load? Shouldn't we be
       // reading these values from the database somehow?
-      set( "og", "og", _og, false );
+      // NOTE: We always calculate this, so lets not cause
+      // a database dirty when loading recipes. This fixes
+      // starting the app and then closing causing a Save Changes.
+      // set( "og", "og", _og, false );
```
rocketman768: Unless someone calls `Recipe::setOg()`, this will cause the og to _never_ get written to the database.

greggree: I'm still digging into the database. But saving the og and fg in the database doesn't make since. Change the yeast, and the fg will change. Change a grain and the OG will change. These two should always be calculated based on other things, that ARE saved in the database. I think we should always calculate these two based on other changes, and not save them in the database. This isssue I'm trying to solve is that when the app starts, it loads a default recipe. When the code calulates the OG/FG for this recipe, and calls set(), the database is dirtied, meaning if I changed nothing and close the app, it asks to save changes. Changes? I didn't do anything.

I'm just getting started with the code, so please take what I say as a question and not religion. I ran the code change, and the recipes still show the right OG and FG. It seems to be calculated all the time.

mikfire: You are digging into parts that usually make my head hurt. I would also
love to see some of it cleaned up.

Historically speaking, our data structures are very closely modelled from
the BeerXML specification. The Recipe BeerXML definition includes an FG and
an OG so our data structures do as well. That's pretty much why we store
them.

Have you tried all the other output formats? Like making sure you can
create properly formatted BeerXML? Do the brewday calculations work? Do the
print functions work?

And please ask questions. I can rarely answer them, but I can usually at
least explain why the code is that way.

Mik


On Tue, May 17, 2016 at 7:20 PM, Greg Greenaae <notifications@github.com>
wrote:

> In src/recipe.cpp
> <https://github.com/Brewtarget/brewtarget/pull/254#discussion_r63621150>:
>
> > @@ -1839,15 +1839,22 @@ void Recipe::recalcOgFg()
> >        // being updated for no good reason.
> >        // NOTE: We are we recalculating all of these on load? Shouldn't we be
> >        // reading these values from the database somehow?
> > -      set( "og", "og", _og, false );
> > +      // NOTE: We always calculate this, so lets not cause
> > +      // a database dirty when loading recipes. This fixes
> > +      // starting the app and then closing causing a Save Changes.
> > +      // set( "og", "og", _og, false );
>
> I'm still digging into the database. But saving the og and fg in the
> database doesn't make since. Change the yeast, and the fg will change.
> Change a grain and the OG will change. These two should always be
> calculated based on other things, that ARE saved in the database. I think
> we should always calculate these two based on other changes, and not save
> them in the database. This isssue I'm trying to solve is that when the app
> starts, it loads a default recipe. When the code calulates the OG/FG for
> this recipe, and calls set(), the database is dirtied, meaning if I changed
> nothing and close the app, it asks to save changes. Changes? I didn't do
> anything.
>
> I'm just getting started with the code, so please take what I say as a
> question and not religion. I ran the code change, and the recipes still
> show the right OG and FG. It seems to be calculated all the time.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/Brewtarget/brewtarget/pull/254/files/92eb6affa051887da1a0ae65f0627b6efcb73652#r63621150>
>



-- 

rocketman768:
> This isssue I'm trying to solve is that when the app starts, it loads a default recipe.

Good, but don't throw out the baby with the bath water. If you never update the database, then you prevent a lot of neat things we can do. For example, you could make a spinoff project that just reads and displays the recipe stats from the database. Mik worked hard on the postgres stuff and I think that is where he's headed.

The real issue you want to fix is that recalcAll() gets called on load. I would approve if you found out why it gets called and prevent it from happening on load.

greggree:
rocketman768 - You are right. I'll move up the stack some and figure out why recalcAll() is being called on recipe load. Now that I got me feet wet working on an open source project, and didn't get called an idiot, I look forward to continuing working with it. I currently use BeerSmith2 and ran into this project from Arch's AUR. Just got done submitting a package for Beersmith2 into the AUR and noticed this project. Installed it and thought that this would work just as good as Beersmtih2 for my Linux driven 10 Gallon brewery and is open source. I love the fact that its Qt and not GTK based. Qt I understand, GTK, not so much.

greggree:
So I've done some more digging. It starts in Database::load, and we call
populateElements(allRecipes, Brewtraget::RECTABLE). This creates a list of
new Recipes. The recipe's constructor is called that sets a bunch of
default values including _og and _fg (set to 1.00) and also sets
_uninitializedCalcs to true. Nowhere in the code does any of the Set* on
the recipe is called. Then we start filling in form fields from a default
recipe. A Recipe::Getter is called, and since _uninitializedCalcs is true,
all the Getters default to calling recalcAll().
The recalcAll() is calling set() on stuff that dirties the database.

As a first attempt, I'm going to try to get these properties from the
database and call Set methods off of recipe during Database::load(). Then
set _uninitializedCalcs to false. I'll have to do some research into the
current default recipes to verify all "Calculated Getters" have values in
the database. One bug I can see is that one of the "Calculated Getters"
doesnt have a value in the database, and since I just set
uninitializedCalcs to false, the value would be empty.

Tying the BeerXML into the recipe seems to break KISS. Everything from the
recipe should come out of the database and not be calculated. Calculations
should only happen on edits.

PR #256, opened 2016-05-21
Closes #255 - Adjusting volume on preboil tab

### LINKED ISSUE #255: Adjusting volume on preboil tab doesn't change bk efficiency
Efficiency into the boil kettle is a factor of two things: density and volume. We are correctly modifying when the density is changed, but the changing the volume doesn't.

mikfire:
Somewhat simple fix. Basically, when I did the scales for input, I changed the
symantics of what the "force" flag means in BtLineEdit. I just didn't change
everything.

There's a signal that needs to be emitted when the field changes, but I only
want it emitted when the user enters a new value in the field. The previous
logic used the forced flag, which was clever but wrong. When I changed the
symantics, I should have changed the flag. This commit fixes that.

And a new rule. Don't use one flag to do two different things. Use different
flags.

### DIFF
--- a/src/BtLineEdit.cpp
+++ b/src/BtLineEdit.cpp
@@ -57,6 +57,7 @@ void BtLineEdit::lineChanged(Unit::unitDisplay oldUnit, Unit::unitScale oldScale
    QString amt;
    bool force = Brewtarget::hasUnits(text());
    bool ok = false;
+   bool wasChanged = sender() == this;
 
    // editingFinished happens on focus being lost, regardless of anything
    // being changed. I am hoping this short circuits properly and we do
@@ -99,8 +100,7 @@ void BtLineEdit::lineChanged(Unit::unitDisplay oldUnit, Unit::unitScale oldScale
    }
    QLineEdit::setText(amt);
 
-   if ( ! force )
-   {
+   if ( wasChanged ) {
       emit textModified();
    }
 }


src/BtLineEdit.cpp
```diff
@@ -57,6 +57,7 @@ void BtLineEdit::lineChanged(Unit::unitDisplay oldUnit, Unit::unitScale oldScale
    QString amt;
    bool force = Brewtarget::hasUnits(text());
    bool ok = false;
+   bool wasChanged = sender() == this;
```
rocketman768: :question: `sender() == this` implies `wasChanged`

mikfire: tl;dr is that yes, it does :)

This slot is called in two possible ways: somebody doing a unit/scale change on a BtLabel, or the BtLineEdit generating an editingFinished signal. The first is connected when constructing screens, the second is done in the constructor for a BtLineEdit.

If the signal is generated by the BtLineEdit it must be because the field was modified by the user. Mostly. See, editingFinished is also generated every time the field loses focus. We actually end up generating a *lot* of editingFinished signals. So the test about four lines further down makes sure something actually changed. If nothing has changed, we just return. This cuts down on a lot of database interactions.

Since we only want to generate the textModified signal when text has actually been modified, the wasChanged flag and the test just a bit further down means the signal will only be emitted if something really changed in the input field.

rocketman768:
:+1:

PR #257, opened 2016-05-21
Closes #251 - Bad css on recipe output

### LINKED ISSUE #251: Bad css on recipe output
Somewhere, I changed the CSS string behaviour in RecipeFormatter, but only half way. If you do a roll over and load the tooltip CSS and then print a recipe, the recipe will be printed with the tooltip CSS.

I've a fix ready to submit.

mikfire:
I removed the cssName variable and simply sent the desired css resource in the
call to getCss(). I also found a small error in the misc table where we
weren't open the table row, which caused weird output.


rocketman768:
:+1:

PR #260, opened 2016-05-23
Fixes #247 -- Database vanished

### LINKED ISSUE #247: Database VANISHED!!!
Everything is GONE, and i'm feking DEVESTATED.
I opened Brewtarget just now, and EVERYTHING is gone.
My recipes, my hop information, my malts, my yeasts…
devestated

OSX 10.8.5
Lastest Brewtarget install.

mikfire:
This doesn't fix the core problem. That comes next, because I think I can.
This implements automatic, under the cover backups.

The OptionDialog has been significantly re-arranged. The user can now select
how often (aka, frequency) to backup, and how many old copies to keep.
Frequency is defined by the number of times a user closes brewtarget and accepts
the changes. Frequency defaults to 4, the number defaults to 10. A frequency
of 1 means always backup on close. If the number to keep is -1, it means never
delete; 0 means never backup; >0 means that number of backups

When brewtarget is closed, we get a counter from the properties file. If that
counter % frequency is not 0, we simply store the updated counter and exit.

If the number to save is 0, we exit.

If we get this far, we pull the backup directory name and the list of existing
files from the properties file.

If counter % frequency is 0, we generate a filename like bt_database.yyyyMMdd.
We then check to see if that file exists in the backup directory. If it does,
we go into a loop appending 0001, 0002 ... until we get a filename that
doesn't exist. If we get to 10000, we give up and just use
bt_database.yyyyMMdd.

Once we have the name, we call backupToDir() which already existed. That just
copies the file for us. We then append the new name to the list.

The final step is to clean up the old copies if we need to. While the size of
the list is greater than the number of copies to keep, we pop the first (aka,
oldest) file name off the list, and delete it. Using a while loop allows the
user to adjust the number of preserved copies down and brewtarget will clean
up after itself.

Finally, we store the new list of files names back into the properties file
and exit.


rocketman768:
>A frequency of 1 means always backup on close. If the number to keep is -1, it means never delete; 0 means never backup; >0 means that number of backups

Is this in a nearby tooltip (especially the -1 part)?

mikfire:
Probably not :)

I will fix.

On Thu, May 26, 2016 at 12:54 AM, Philip Lee <notifications@github.com>
wrote:

> A frequency of 1 means always backup on close. If the number to keep is
> -1, it means never delete; 0 means never backup; >0 means that number of
> backups
>
> Is this in a nearby tooltip (especially the -1 part)?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/Brewtarget/brewtarget/pull/260#issuecomment-221777194>
>



-- 

mikfire:
Ok. This is now a serious merge candidate. 

I have included the best fixes I can figure to make sure we don't run two brewtargets. I have removed the pid file stuff. This needs to be tested on more than just linux, but test with caution. If I got it wrong, your database will be removed.

I have removed all concept of the temporary backup. No more dirty bits, no more prompts to save or restore/rollback. All writes are made directly to the db.

And I added the tooltip, as was requested.

rocketman768:
:+1: Looks good to me

mikfire:
There's something weird in there. Suddenly, I have a database in my source/build directory. I need to trace that down before I complete the merge.

mikfire:
That wasn't caused by any work in this tree. I'm merging.

There is a lot to test before v2.4 is ready.

PR #262, opened 2016-05-29
Fixes #270: avoid recipe recalc on load

### LINKED ISSUE #270: Avoid recipe recalc on load
Don't do this. It's bad mkay.

greggree:
Since the recipes are not loading the calculated values from the database, we can use the _uninitializedCalcs bool to determine if we need to call set or emit. If the recipe is just being loaded, _uninitializedCalcs will be true. After the initial load, it will be set to false. If it is false, call set and or emit  for changes. This stops the database getting dirty when no changes are made, and is a minor code change. And if something is changed, the database will then get dirtied, I'm still working on getting the values from the database during loading, but haven't found the right location yet. Since the same functions are used to set the database values in the first place, they should match anyways.

Fixing should be faster now the I got the project loading and running in QtCreator. A lot easier to check values and properties out since qtcreator knows about QtObjects. ddd just shows pointer addresses and makes it hard to figure out what the pointer is pointing too.

### DIFF
--- a/src/recipe.cpp
+++ b/src/recipe.cpp
@@ -1373,6 +1373,8 @@ void Recipe::recalcAll()
    // Infinite recursion possible, since these methods will emit changed(),
    // causing other objects to call finalVolume_l() for example, which may
    // cause another call to recalcAll() and so on.
+   //
+   // GSG: Now only emit when _uninitializedCalcs is true, which helps some.
 
    // Someone has already called this function back in the call stack, so return to avoid recursion.
    if( !_recalcMutex.tryLock() )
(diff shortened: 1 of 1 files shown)


rocketman768:
I like this solution. Does it have any side effects?

greggree:
Not that I have found. Been testing and running into other bugs that I am trying to log and debug. Recipes look good, and unless you make a change, when you close the app, there is no save request.

rocketman768:
:+1:

mikfire:
He signed off. I'm easy.

PR #268, opened 2016-05-30
Closes #267 - Moving mash steps doesn't update form.

### LINKED ISSUE #267: Moving mash steps doesn't update form.
After the fix for #265, I noticed that the UI isn't updating to show the mash steps being moved around.

mikfire:
That was .. a mess. I don't know how far back this error goes, but I think its
old.

First, the SQL statement in Database::swapMashStepOrder was wrong. As
@greggree noticed, it was using msid instead of mash_id. The update itself
changed the stepNumber to be the id of the steps, which ... well, just wasn't
right. So I fixed that, and reworked the QString a little.

Second, there was nothing in the MashStepTableModel to do the right thing when
the stepNumber changed signal was received. I added a method that just removes
the rows and re-adds them. I do not disconnect any signals, which may come to
haunt us later.

I think this fixes the issues.

### DIFF
--- a/src/MashStepTableModel.cpp
+++ b/src/MashStepTableModel.cpp
@@ -86,6 +86,23 @@ void MashStepTableModel::setMash( Mash* m )
    }
 }
 
+void MashStepTableModel::reorderMashSteps()
+{
+   if( mashObs && steps.size() > 0)
+   {
+      // I am NOT disconnecting the signals. I just want things redrawn
+      beginRemoveRows( QModelIndex(), 0, steps.size()-1 );
+      // Remove mashObs and all steps.
+      steps.clear();
+      endRemoveRows();
+
+      QList<MashStep*> tmpSteps = mashObs->mashSteps();
+      beginInsertRows( QModelIndex(), 0, tmpSteps.size()-1 );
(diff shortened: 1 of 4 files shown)


src/MashStepTableModel.cpp
```diff
@@ -86,6 +86,23 @@ void MashStepTableModel::setMash( Mash* m )
...
+      // I am NOT disconnecting the signals. I just want things redrawn
+      beginRemoveRows( QModelIndex(), 0, steps.size()-1 );
+      // Remove mashObs and all steps.
+      steps.clear();
+      endRemoveRows();
+
+      QList<MashStep*> tmpSteps = mashObs->mashSteps();
+      beginInsertRows( QModelIndex(), 0, tmpSteps.size()-1 );
+      steps = tmpSteps;
+      endInsertRows();
```
rocketman768: Instead of removing and adding the steps back, could you just call

    beginResetModel();
    endResetModel();

?

mikfire:
Sort of, but not really. I like this idea better -- we just move the rows around that need moving around.

rocketman768:
Then, :+1:

PR #278, opened 2016-06-12
Fixes #88: high dpi printing

### LINKED ISSUE #88: Printed pages do not fill page width on high-dpi displays

rocketman768:
Really, this fixes a lot more than #88. It fixes a lot of html/css compatibility issues that were screwing up the rendering of the html in QTextBrowser and the printout. Something is still adding a large border to the printout version, but I will leave that for another bug fix.

## Relevant Links

 - http://doc.qt.io/qt-5/richtext-html-subset.html
 - https://validator.w3.org/check
 - https://www.sitepoint.com/understanding-and-using-rem-units-in-css/
 - https://www.w3.org/TR/1999/REC-html401-19991224/
 - http://www.freeformatter.com/html-formatter.html

### DIFF
--- a/css/brewday.css
+++ b/css/brewday.css
@@ -1,18 +1,21 @@
+html {
+   font-size: 100%;
+}
 body{
-   width: 8.5in;
-   margin: auto auto auto 10px; 
+   width: 90%;
+   margin: auto auto auto auto;
 }
 caption { caption-side:top; }
 ul { list-style-type:circle; }
 h1 
 { 
-   font-size:36px;
+   font-size:3rem;
    font-weight:bold;
    text-align:left;
 }
 h2
 {
-   font-size:20px;
+   font-size:2rem;
    font-weight:bold;
    text-align:center;
    border-bottom-style:solid;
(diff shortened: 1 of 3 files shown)


rocketman768:
## Print Preview

### Before

before

### After

after

## Print

### Before

print_before

### After

Something is off with the page margins here, so we need to track down what is setting the margins to reclaim some more space in another pull request.

print_after

mikfire:
Be careful with redoing the margins on the prints. It was specifically requested that we leave a margin so that people could use a hole punch and put them in a three ring binder.

rocketman768:
> Be careful with redoing the margins on the prints. It was specifically requested that we leave a margin so that people could use a hole punch and put them in a three ring binder.

OK. I'll test for that.

> And you want line 3 of recipe.css for the margins.

It's not the body margins. It's something external to the html and css, because the print preview (see above) and the exported html viewed in a browser look correct. I tried setting the QPrinter margins to 0 explicitly, but no dice.

PR #286, opened 2016-06-20
Script to make it easy to update the default db with new ingredients

rocketman768:
It sucks directly editing default_db.sqlite. Just make a big .json file and let the script import the data into the database.


mikfire:
So why don't we just pull this all from brewerwall?

PR #293, opened 2016-07-06
Closes #291 - Segfault merging database

### LINKED ISSUE #291: Segfault merging database
When updating from 2.3.1 to current develop branch (matches 2.4 branch), I get a popup "There may be new ingredients and recipes available. Would you like to add these to your database?". If I choose yes, brewtarget segfaults with this trace:
```
#0  Database::copy<Yeast> (this=this@entry=0x2450510, object=0x1, displayed=displayed@entry=true, keyHash=keyHash@entry=0x2450580)
    at /tmp/brewtarget/src/brewtarget/src/database.h:790
#1  0x0000000000530acd in Database::newYeast (this=0x2450510, other=<optimized out>)
    at /tmp/brewtarget/src/brewtarget/src/database.cpp:1509
#2  0x0000000000525cde in Database::updateDatabase (this=this@entry=0x2450510, filename=...)
    at /tmp/brewtarget/src/brewtarget/src/database.cpp:4926
#3  0x00000000005456b5 in Database::load (this=this@entry=0x2450510)
    at /tmp/brewtarget/src/brewtarget/src/database.cpp:351
#4  0x0000000000545871 in Database::Database (this=0x2450510)
    at /tmp/brewtarget/src/brewtarget/src/database.cpp:106
#5  0x00000000005459d5 in Database::instance ()
    at /tmp/brewtarget/src/brewtarget/src/database.cpp:654
#6  0x0000000000509d89 in Brewtarget::initialize (userDirectory=...)
    at /tmp/brewtarget/src/brewtarget/src/brewtarget.cpp:435
#7  0x000000000050ad47 in Brewtarget::run (userDirectory=...)
    at /tmp/brewtarget/src/brewtarget/src/brewtarget.cpp:522
#8  0x00000000004c107e in main (argc=<optimized out>, argv=<optimized out>)
    at /tmp/brewtarget/src/brewtarget/src/main.cpp:70

```

mikfire:
In all of my wonderful refactors, I missed the function pointers in
Database::makeTableParams(). The function pointers resulted in newYeast (for
example) being called with 0x1 as the pointer to other, which lead to a
sigsegv.

I've modified the function pointers so they do the approximate right thing,
and no cores. I also had to clean up a small problem with an insert statement.


rocketman768:
I think this can be simplified. I will make some in-line comments below.

src/database.h
```diff
@@ -68,7 +68,7 @@ typedef struct
 {
    QString tableName; // Name of the table.
    QStringList propName; // List of BeerXML column names.
-   BeerXMLElement* (Database::*newElement)(void); // Function to make a new ingredient in this table.
+   BeerXMLElement* (Database::*newElement)(int); // Function to make a new ingredient in this table.
```
rocketman768: Since all the functions take a pointer instead of an `int`:

    BeerXMLElement* (Database::*newElement)(BeerXMLElement*);

src/database.cpp
```diff
@@ -4690,7 +4690,7 @@ Yeast* Database::yeastFromXml( QDomNode const& node, Recipe* parent )
 
 QList<TableParams> Database::makeTableParams()
 {
-   typedef BeerXMLElement* (Database::*NewIngFunc)(void);
+   typedef BeerXMLElement* (Database::*NewIngFunc)(int);
```
rocketman768: BeerXMLElement* (Database::*newElement)(BeerXMLElement*);

src/database.cpp
```diff
@@ -4705,7 +4705,7 @@ QList<TableParams> Database::makeTableParams()
       "lauter_deadspace" << "top_up_kettle" << "hop_utilization" <<
       "notes";
    tmp.newElement =
-      (NewIngFunc) (Equipment*(Database::*)(void))
+      (NewIngFunc) (Equipment*(Database::*)())
```
rocketman768: Leave the `(NewIngFunc)` cast and remove the `(Equipment*(Database::*)())` cast.

src/database.cpp
```diff
@@ -4719,7 +4719,7 @@ QList<TableParams> Database::makeTableParams()
       "max_in_batch" << "recommend_mash" << "ibu_gal_per_lb";
    tmp.newElement =
       (NewIngFunc)
-      (Fermentable*(Database::*)(void))
+      (Fermentable*(Database::*)())
```
rocketman768: Leave the `(NewIngFunc)` cast only...ditto for the rest.

rocketman768:
You know, I really hate that we have to cast a function pointer. It really forces the compiler to shut up about a very dangerous thing. This could be a good reason to start using c++11 syntax with a lambda to wrap up everything into a parameterless call without unsafe casting. That way, if the function signatures change, the coder at least gets an error message.

    typedef struct {
        ...
       std::function<void(void)> newElement;
    } TableParams;
    ...
    tmp.newElement = [&]() {
        this->newYeast(0);
    };
    ...
    oldid = tp.newElement()->_key;

mikfire:
I know little about the lambda fun, but I hate the function pointers. I
like that syntax, and would be more than happy to do the grunt work to put
it everywhere.

Mik

On Wed, Jul 6, 2016 at 10:47 PM, Philip Lee <notifications@github.com>
wrote:

> You know, I really hate that we have to cast a function pointer. It really
> forces the compiler to shut up about a very dangerous thing. This could be
> a good reason to start using c++11 syntax with a lambda to wrap up
> everything into a parameterless call without unsafe casting. That way, if
> the function signatures change, the coder at least gets an error message.
>
> typedef struct {
>     ...
>    std::function<void(void)> newElement;
> } TableParams;
> ...
> tmp.newElement = [&]() {
>     this->newYeast(0);
> };
> ...
> oldid = tp.newElement()->_key;
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/Brewtarget/brewtarget/pull/293#issuecomment-230965325>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AKxgF0_tlYBR8w_qW8YxpDei9SDLp_Ihks5qTGingaJpZM4JFpwN>
> .
>



-- 

mikfire:
I think that is much better. I can actually read what it does without using stack exchange. I've explained pretty thoroughly what I did in the commit.

CMakeLists.txt
```diff
@@ -87,6 +87,8 @@ IF( ${NO_QTMULTIMEDIA} )
    ADD_DEFINITIONS( -DNO_QTMULTIMEDIA )
 ENDIF()
 
+ADD_DEFINITIONS("-std=c++11")
```
rocketman768: I think you aren't adding a precompiler definition here, and modifying `CMAKE_CXX_FLAGS` with this option will only work with gcc/clang. I think we want to do [this](https://crascit.com/2015/03/28/enabling-cxx11-in-cmake/)

    set(CMAKE_CXX_STANDARD 11)
    set(CMAKE_CXX_STANDARD_REQUIRED ON)

rocketman768:
If you fix the cmake thing, I'm good with the rest of this.

mikfire:
That's interesting. Why did the clang build fall over?

rocketman768:
Looks like that variable was introduced in make 3.1. Time to bump the standards version from 2.8.8 to 3.1, and I think you can eliminate the `CMAKE_POLICY` line unless something complains at you.

PR #296, opened 2016-07-10
Closes #282 -- batch sparge type has no affect on water slider amount…

### LINKED ISSUE #282: [Mash Designer] batch sparge type has no affect on water slider amount range
@mattipt says:

Selecting Batch Sparge as the step type has no effect on the range of the amount slider (used to work in 2.3 I believe.)

mikfire:
… range.

This was definitely part of the batch sparge type stuff. I didn't call updateMaxAmt() when the type was changed to batch or when it was changed back to infusion.

It also drove me nuts that the field wouldn't change until after you entered the temp. So I fixed that too.


rocketman768:
:+1:

PR #297, opened 2016-07-10
Closes #283 -- temperature slider reversal

### LINKED ISSUE #283: [Mash Designer] temperature slider reversal
@mattipt says:

If hitting the target temperature would result in water hotter than boiling, the temperature slider reverses, with a higher-than-boiling temperature as its minimum value and boiling point as the maximum.

mikfire:
The slider didn't really reverse. Its behavior simply didn't make sense.

The max temp label is taken either as the configured boiling point of water (bpow, for short) or 100C. In other words, it is treated mostly like a constant and not calculated.

The min temp label, on the other hand, is calculated. We never bothered to check if the results made any sense. While I kind of see the point of staying true to the maths, I think we should consider physics too.

I have put a number of checks in place to make sure if the temperature exceeds the configured bpow, the temperature is set to the bpow. The min/max labels will not exceed the bpow, the temperature label will not, the step starting or target temps will not either.

### DIFF
--- a/src/MashDesigner.cpp
+++ b/src/MashDesigner.cpp
@@ -97,7 +97,6 @@ void MashDesigner::show()
       );
       return;
    }
-
    setVisible(nextStep(0));
 }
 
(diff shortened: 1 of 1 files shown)


rocketman768:
:+1: 

Can we also make test cases for this kind of stuff?

PR #305, opened 2016-08-22
Add Gladfield Malts range to default database.

bcbnz:
Gladfield Malts are a Dunsandel, New Zealand based maltser. This pull request.
adds their current range of malts to the default database.

The data is extracted from the BeerSmith download they provide at
http://www.gladfieldmalt.co.nz/our-malts/.


rocketman768:
👎 missing corresponding entries in `bt_fermentable` table.

    $ ./scripts/sqlitediff.sh data/default_db.sqlite master origin/pull/305

[db_a4912b1.diff.zip](https://github.com/Brewtarget/brewtarget/files/429391/db_a4912b1.diff.zip)

bcbnz:
Sorry! I've pushed a new (squashed) commit to the branch which should take care of that, providing I've got the right idea about how the entries should work... I've only just started using BrewTarget and it seemed to work without adding the bt_fermentable entries on my local copy so I wasn't aware of its existence.

mikfire:
@rocketman768 is this good to merge? I am going to be doing the translations tonight, and want to get the last of this merged in

rocketman768:
👍

Issue #319, opened 2016-09-15
Initialising a new database


mordante:
I ran into an issue while creating a new database with a new user account. The problem is the application does not start when creating a new database, with a SQL error. I've seen the issue before but it is rather elusive to reproduce. It seems to have to do with SQL escaping. I think I have a solution, but since reproducing the issue is hard, it's hard to validate the fix. I expect to have a proper pull request next weekend. (If I fail to find a proper fix I'll give more details on the bug.)

I consider this issue a blocker for 2.4.0, but it does not affect translatable strings.

mikfire:
The two referenced pull requests have been merged. Can this issue be closed?

mordante:
Yes I can't recall other issues, else I'll file a new bug report or pull request.

PR #321, opened 2016-09-17
Improve handling of fatal exceptions.

mordante:
While working on issue #319 the application terminated with the following
error message on the console:
terminate called after throwing an instance of 'QString'

Instead of logging this message on the console show a message box to the
user with the contests of the QString thrown. Do the same for
std::exceptions, and a generic message if the exception is of another
type.


mikfire:
That looks good. I will merge this soon.

PR #334, opened 2016-11-22
Russian translation

naudhr:
A bit of Russian translation strings.

### DIFF
--- a/translations/bt_ru.ts
+++ b/translations/bt_ru.ts
@@ -112,7 +112,7 @@
     </message>
     <message>
         <source>Estimated calories (per 33 cl)</source>
-        <translation type="unfinished"></translation>
+        <translation>Приблизительно калорий (на 0,33л)</translation>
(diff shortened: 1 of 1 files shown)


dpettersson:
@rocketman768  How do we handle translations? Do we have anyone that can proof read this?

rocketman768:
@dpettersson We previously did not appoint people to be in-charge of each language. I think that would be a very good change to make to manage some chaos. This one mostly just looks like filling in unfinished translations and missing punctuation, so it's probably fine.

rocketman768:
I didn't answer your question about how we handle translations. Usually we get translations from Launchpad. See [translations/README](https://github.com/Brewtarget/brewtarget/blob/develop/translations/README). I think to incorporate the changes in this pull request, we'd have to convert this .ts file to a .po file, upload to Launchpad, and make sure we don't overwrite the existing translations there.

Issue #158, opened 2016-01-29
Initial value of IBU Adjustments erronous on system use coma as decimal separator


theophae:
After a fresh install of BT 2.3.0 on windows, I noticed that the default values of IBU Adjustments were set to 10 000,000 instead of 10,000. This is obviously due to the use of coma as separator in my system configuration.
This bug is not blocking as if I set 10,000 as a new value it will remains as is.

pricelessbrewing:
Tested on latest dev build, looks like it's fixed. Close?

pricelessbrewing:
@dpettersson can you test this on linux and see if it's an issue for you?

dpettersson:
Sure, i can test it.
But i saw something earlier. The develop branch has (compared to 2.4)  141 more commits and there are 2 that doesn't exist in develop that are in 2.4. Maybe we should merge develop into 2.4 (or vice versa). It would give a great boost to the functionality if we could do something like that.

dpettersson:
It seems to work for me.
I tested it like this:
I moved the sqlite-db and started the program. It looked OK.

mattiasmaahl:
closed?

matty0ung:
Since, from the most recent comments, it looks like this is fixed, I'm closing.  Please re-open if I've misunderstood.

PR #336, opened 2017-01-12
Possibly closes #166 and #158

### LINKED ISSUE #166: Problems with OGs and FGs for each style (2.3.0)
I installed the Linux 64-bit version of the 2.3.0, and I'm facing some problems with the OGs and FGs for each style. They are appearing with dots and commas mixed, and higher values. 
For example: Belgian Golden Strong Ale
OG min: 1,070 sg
OG max: 1.095,000 sg
FG min: 1.050,000 sg
FG max: 1.016,000 sg

Do you know what's wrong?

Thanks

mikfire:
This is a hard thing.

The essence is that in BtLineEdit::setText(BeerXMLElement *element), the conversion went from QVariant to QString to double. The problem was that the QString ended up as something like "1.045". The conversion from QString to double is locale aware, so en_US would get 1.045 but locales like bt_PR would get 1e3+45.

I fixed this by replacing the QVariant->QString->double conversion with a QVariant->double conversion. It *seems* to work -- I've tried this with en_US, bt_PR, de_DE and fr_FR and they all *appear* to do the right thing.

I don't think we need to fix anything else. The values written to the database are written as doubles, so this kind of conversion is irrelevant. On the other hand, I have demonstrated an annoying inability to test my own code and my database my not behave like anybody else.

### DIFF
--- a/src/BtLineEdit.cpp
+++ b/src/BtLineEdit.cpp
@@ -41,7 +41,7 @@ BtLineEdit::BtLineEdit(QWidget *parent, Unit::UnitType type) :
    _forceUnit = Unit::noUnit;
    _forceScale = Unit::noScale;
    */
-   
+
    connect(this,SIGNAL(editingFinished()),this,SLOT(lineChanged()));
 }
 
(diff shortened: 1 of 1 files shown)


malavv:
As I mention, I wasn't able to replicate on my side so I won't be of great help on this.

However, the code seems alright; lots of formatting changes though. The main function you modified looks better like this. The string middle conversion was a bit weird.

As a side comment, is there a specific reason we use _toLatin1_ instead of _toUtf8_ even though I realize that this will logically only be numbers and a few punctuation.

mikfire:
Oh. The only way I could reliably exercise the bug was to say: LC_ALL=pt_BR ./src/brewtarget

I could sometimes see the problem using de_DE or fr_FR, but it was oddly random.

PR #343, opened 2017-01-24
Limit max infusion amount in Mash Designer by target boil size

### LINKED ISSUE #339: Mash Designer lets you over-shoot target collected wort volume
Common workaround for BIAB is to use the Mash Designer dialog to create a full-volume infusion. Currently the sliders have an upper limit only based on the size of the mash tun, so it takes some careful nudging to get exactly the target collected wort. Can't think of a use case where you would want to collect more than the target volume, so the sliders should hit a limit at exactly 100%.

I think this is a simple enough change that I can tackle it.

mash-overfill

Daedalus12:
When using the Mash Designer, the upper limit of the range for the "Infusion Amount" and "Infusion Temp" sliders is now restricted by the collected wort vs. target boil size as well as the mash tun fullness.

### DIFF
--- a/src/MashDesigner.cpp
+++ b/src/MashDesigner.cpp
@@ -213,18 +213,24 @@ double MashDesigner::minAmt_l()
 // However much more we can add at this step.
 double MashDesigner::maxAmt_l()
 {
+   double amt = 0;
+
    if ( equip == 0 )
-      return 0;
+      return amt;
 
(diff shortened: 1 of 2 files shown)


Daedalus12:
Ah, just realized that this doesn't account for top-off water in the equipment profile. Will fix.

PR #347, opened 2017-01-27
Upgrade signal/slot connect calls to Qt5 style connect

brianrower:
See document here on newer signal/slot connect style: https://wiki.qt.io/New_Signal_Slot_Syntax
Biggest benefit is compile time errors when a signal or slot no longer exists, was typo'd, etc.
This is all the ones that were EASY to convert, the new style doesn't like functions with overrides (same function name different parameters) so I skipped those for now.
No functional changes, just code refactoring.


mikfire:
Oh. I've been thinking of doing this for a while. I am slightly jealous that you got it done first. 

The harder ones to switch were almost always written because the older style was doing string compares and it was the only way I could figure out to do it. It should be possible -- with a bit of casting and maybe a lambda or two -- to get those collapsed down to a single method.

If you can fix the one conflict while I review the code, we can get this merged in

brianrower:
Fixed conflict

PR #353, opened 2017-02-18
Cleanup on Fermentable editing

brianrower:
This is the first of a lot of refactoring/cleanup I'd like to do.
This changes BeerXMLElement to not have a default constructor, it now requires a table and key.
Doing this allows a BeerXMLElement to gather it's values from the database in it's constructor.
The next change is the Fermentable object now gathers all it's values from the database in the constructor and stores them in member variables. the getter functions now just return the member variable, the setters no longer write to the DB, instead the save() function has to be called on the Fermentable which commits everything to the database all as 1 SQL statement. This is much more efficient in the FermentableEditor scenario, but a little less efficient in the FermentableTableModel scenario where a user can actually change and save 1 column of the fermentable, the save function could be optimized to only save what has been changed if people don't like this. I've also started creating static const QString's for strings that are used multiple times as this is common programming best practice.

### DIFF
--- a/src/BeerXMLElement.cpp
+++ b/src/BeerXMLElement.cpp
@@ -27,6 +27,9 @@
 #include "brewtarget.h"
 #include "database.h"
 
+static const QString kFolder("folder");
+static const QString kName("name");
+
 BeerXMLElement::BeerXMLElement(Brewtarget::DBTable table, int key)
    : QObject(0),
      _key(key),
@@ -84,14 +87,14 @@ void BeerXMLElement::setDisplay(bool var)
 QString BeerXMLElement::folder() const
 {
    if ( _folder.isEmpty() )
-      _folder = get("folder").toString();
+      _folder = get(kFolder).toString();
 
    return _folder;
 }
 
 void BeerXMLElement::setFolder(const QString var, bool signal)
 {
-   set( "folder", "folder", var );
+   set( kFolder, kFolder, var );
    _folder = var;
    if ( signal )
       emit changedFolder(var);
@@ -107,7 +110,7 @@ QString BeerXMLElement::name() const
 
 void BeerXMLElement::setName(const QString var)
 {
-   set( "name", "name", var );
+   set( kName, kName, var );
    _name = var;
    emit changedName(var);
 }
@@ -258,9 +261,19 @@ void BeerXMLElement::set( const char* prop_name, const char* col_name, QVariant 
    }
 }
 
+void BeerXMLElement::set(const QString &prop_name, const QString &col_name, const QVariant &value, bool notify)
+{
+   set(prop_name.toUtf8().constData(), col_name.toUtf8().constData(), value, notify);
+}
+
 QVariant BeerXMLElement::get( const char* col_name ) const
 {
    return Database::instance().get( _table, _key, col_name );
+}
+
+QVariant BeerXMLElement::get( const QString& col_name ) const
+{
+   return get(col_name.toUtf8().constData());
 }
 
 void BeerXMLElement::setInventory( const char* prop_name, const char* col_name, QVariant const& value, bool notify )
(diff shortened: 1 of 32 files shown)


brianrower:
Oh looks unit tests, guess I should try running those, closing until I look into that

brianrower:
original submit needed a connection and some new save() calls in the tests, working now

mikfire:
Hm. We very intentionally don't have member variables. This was part of the design choices we made when transitioning to the SQL backend. I am a little hesitant to change that design without further discussion.

I hacked in a bit of caching for name, display and deleted to address some performance issues but my investigations at the time indicated any further caching didn't significantly impact the performance. I was only looking at load times, and would have missed any later issues.

So a few questions. Are you making an argument that we should do this for all of the BeerXML classes, and this is just the first of many? Are you making an argument that there is something special about fermentables that requires making a very large exception, and if you are what is it? Have you tested error cases, like crashes half way through a save()

I will need some time to read over the code and understand what's going on. I am not ignoring this, but my life is kinda busy right now so it will take some time.

brianrower:
Yes, I'm absolutely making the argument that we should do this for all BeerXML classes, nothing special about fermentables, just happened to be the first I picked to use as an example. The save() is 1 SQL statement...so if you crash before the same finishes you will not have your changes, if you crash after, you will have your changes. I'm certainly open to (and expected) discussion, that's why I did it to one class before doing it to all.

Here's WHY I think this should be done to all BeerXML:

1. Reduce Over signaling - Prior to this change, if you were to modify a fermentable using the fermentable editor dialog, when you closed the dialog, each column (there's about 16 in this case) would run a SQL update and each column would send a changed() signal. This changed signal is sent by the DB, so essentially the whole world is listening to it and then has to determine if it cares about that signal or not (by checking what type the sender is and what property was changed). So if you were to want to debug Recipe::recalcAll(), and you saved a fermentable, you would get 16 calls to recalcAll when the save occurred, and you only cared about the last one. To reduce this signaling, you must have a signal that happens once rather than happening 16 times, hence saving all values in members and saving as 1 DB query.

2. Create Transparent Signal Connecting - In the world I work in, it's considered a bad pattern to have other objects connect you to a signal. For example, if you read through Recipe.cpp you'd think it was listening for almost NO signals, but then you look into Database.cpp and realize Database is connecting it to the changed() signals of almost everything. This goes against the principles of C++/object oriented design to encapsulate related code into a class. Here you've actually split the related code between Recipe and Database

3. Use inheritance constructs instead of template functions - The DB template functions are confusing and in my opinion also go against the OOO design principles. Instead these functions could be moved into the BeerXML inheritance tree which should be much more clear what they're doing and should even reduce code size a bit (remember template functions are just copied by the compiler). This COULD be done without member variables, but I assume you it makes it much easier if the member variables are there.

4. Performance: I agree with you that on a desktop computer, there's no noticeable performance change. It's not like the recalc function is doing a 5,000 variable linear regression, it's pretty simple calculations, so doing them 16 times still takes milliseconds, but....I've reduced 16 DB queries into 1 and 16 recalcs to 1....you may not notice it, but there's a performance gain. One of the idea's I've been kicking around (after a long list of refactoring) is a QML frontend w/ Qt for Android. I'm guessing Android you might have noticed a lag with 16 sqlite queries, but then again, maybe not.

mikfire:
Thanks for the detailed response. Those are ... uhh ... good arguments.

I like the ideas and I am willing to merge, as soon as I've had a chance to look at the changes and understand where this is headed. This is a significant change in design philosophy and, since I am only a pro tem bdfl, I would also like to give @rocketman768 a chance to weigh in.

brianrower:
Yeah take your time, obviously you guys know the code base better than I do, so hopefully you'll understand the implications better than I.

mikfire:
I like this. It looks clean and enhances the readability of the code significantly. 

I am curious as to why you've replaced the reliance on the Qt metaobject system -- specifically, the objectclass name -- with a method that returns a static string. I don't mind, and will merge the change regardless of the answer. But I can't learn if I don't ask.

brianrower:
The metaobject reliance requires an instance of an object to get that value. If you look at the change in Database::newIngredients (template function) you'll see I moved the new T() to later in the function (after the key is known). To enforce the new BeerXMLElement's constructor (now that there's no default constructor) you must have the key before creating the object, but w/ the metaobject setup, you couldn't get the table to insert into before creating an object of that type...chicken and egg issue.

Issue #94, opened 2015-09-27
Mash designer - lower step temperature causes negative infusion


rocketman768:
If the current step temperature is less than the current one, you get negative infusion volumes available to you.

![negative_infusion](https://cloud.githubusercontent.com/assets/752558/10124043/26de2ca6-64ff-11e5-8f84-e97e38bb8405.png)

rocketman768:
This problem is bigger than I anticipated. The mash designer is totally broken when the current target temp is lower than the previous target temp. The problem is that the functions `MashDesigner::maxTemp_c()`, `MashDesigner::minTemp_c()`, `MashDesigner::minAmt_l()`, and `MashDesigner::maxAmt_l()` are all implicitly assuming the temperature is going up.

I am not confident that I can fix this without a significant risk of regression for 2.2.0. Will push this off until next release, as the risk is high, and this is an unlikely use case.

f1oki:
Wouldn't it suffice to use something like

```C++
if (currentTargetTemp > previousTargetTemp) { ... }
else { ... }
```

in `MashDesigner::maxTemp_c()` and  `MashDesigner::minTemp_c()`?

rocketman768:
I tried it. The problem with that is that the sliders need to move in the opposite direction when the temperature goes down. That requires more logic throughout the rest of the file. It's doable, just not a quick fix.

f1oki:
Right on! I hadn't thought about that.

PR #356, opened 2017-03-04
Mash designer update

kapinga:
Resolves bug #94, along with other functionality improvements to the Mash Designer.

- Updated the logic for the minTemp_c(), maxTemp_c(), and minAmt_l() functions, by checking whether or not the target step temperature represents heating or cooling. Additional changes ensure that the values returned by these functions are sane (minimum values are always less than maximums, and stay within sane bounds - temps range from 0 C to boiling, amounts range from 0 to a full tun/brewpot).

- When the user changes the step type or the target temperature, the sliders are set to sane initial guesses. The first infusion is initially set for ~1 lb/qt (2.05 l/kg) mash thickness; all other infusions are initially set to the minimum amount. Sparges are initially set to the maximum amount available (which should fill the brewpot and/or drained mash tun).

- Finally, if the user requests an impossible to achieve temperature (requiring water above boiling or below freezing), the target temperature is adjusted to be within the bounds of achievable temperatures, using the new maxTargetTemp_c() function.


kapinga:
I forgot to mention the minor change to Algorithms.h so that Polynomials.order() returns an unsigned int instead of size_t, since intPow requires an unsigned int power. This probably could be reverted and/or changed so that intPow uses size_t instead.

kapinga:
So I'm still figuring out how both Github and the standard git branching model work in practice. As far as I can tell, this should have been developed against stable/2.4.0. I am working on that now, but in the meantime I'm closing this PR.

PR #361, opened 2017-03-19
Closes #357 in version 2.3.1

### LINKED ISSUE #357: Mash wizard does not adjust sparge water temperature for changes in mash temperature
Using brewtarget 2.3.0 on Ubuntu:

I create a mash-in step, set a target temperature, and then run the mash wizard and it spits out strike water volume and temperature for that target, and a volume and temperature for the batch sparge step.

If I change the target temperature of the mash and re-run the mash wizard, it adjusts the initial strike water temperature, however does not change anything in the batch sparge step. I get the same behavior if I remove all mash steps and start over, using a different mash target temp. It will spit out the same batch sparge volume and water temperature no matter what temperature my mash is at.

kapinga:
Closes bug #357. The mash wizard calculates the wrong sparge water temperature in most instances.

There was an off-by-one error that caused MashWizard to reference a newly created MashStep when determining the temperature before the sparge. This meant that it always assumed the mash was at the default temperature (67 C) before sparging, and caused the sparge water temperature to be the same regardless of the actual mash temperature prior to sparging.

Since version 2.3.1 has already been released (but not on Windows!), should we create a new 2.3.2 branch for this?

### DIFF
--- a/src/MashWizard.cpp
+++ b/src/MashWizard.cpp
@@ -261,7 +261,7 @@ void MashWizard::wizardry()
          steps.append(mashStep);
       }
 
-      int lastMashStep = steps.size()-1;
+      int lastMashStep = steps.size()-2;
       tf = mash->spargeTemp_c();
       if( lastMashStep >= 0 )
          t1 = steps[lastMashStep]->stepTemp_c() - 10.0; // You will lose about 10C from last step.


kapinga:
FYI, as far as I can tell, the 2.3.1 branch doesn't pass the TravisCI build check. This PR changes exactly1 character and doesn't actually break things.

kapinga:
Closing this PR. #357 is fixed in 2.4.0, and I assume we're not releasing any more updates to the 2.3 line.

Issue #54, opened 2015-06-27
Incorrect balloon comments for Strike Temp and Final Temp


clmoore3rd:
The balloon comments for Strike Temp and Final Temp are incorrect on the Brew It! tab.  This is for the 64bit Linux package brewtarget_2.1.0-1_amd64.deb. 

They read:
Strike Temp "Temperature of mash after dough in"
Final Temp "Temperature of mash before mash out"

According to the online manual the balloon comments should be:
Strike Temp "The temperature of your strike water before dough in"
Final Temp "The temperature of your mash after dough in"

janjachnik:
I think the terms are ambiguous.

It might be clearer to have 3 temperatures:

1) Temperature before dough in (strike water temperature)
2) Temperature after dough in
3) Final mash temp (before mash out)

clmoore3rd:
With those three temperatures it should be possible to calculate actual specific heat of the mash tun, right?  So where the value in the equipment profile is feed forward to calculate strike water temp to get mash-in temp, having an actual value calculated from brewday measurements would give the user feedback to dial in the equipment profile:)

My approach was to iterate the specific heat value so the mash wizard calculator (great feature BTW) closely matched actual results from previous brew sessions.

eltomek:
Are the _Strike temp_ and _Final temp_ fields really needed/usable on the `brewNoteWidget`? The _Strike temp_ value (unless the brewer decided else, but the calculated values are what he uses Brewtarget for) is the one that we have in the _Mash Step'_s _Infuse temp._ field. If he decided to use different strike temp. he could store it simply in the _Notes_ field. I have the same concerns regarding  the _Final temp_ (temp. of mash before mash out). 
Of course, this may be some valuable parameter for the brewer but until it is used to calculate anything in Brewtarget it is a redundant data field that as a result contributes (as a practice) to make software overcomplicated and unfriendly to new users.

mikfire:
The idea of the strike temp field was to allow the user to record the temperature they actually heated the strike water to -- I am sometimes distracted and sometimes (maybe) strike a little hotter than expected. Recording this sort of information is exactly what I wanted the brewnotes for -- so the little deviations that just happen on a brewday can be recorded.

[...]

PR #363, opened 2017-03-23
Closes bug #54

kapinga:
Bug #54 started with an inconsistency on some tooltip labels between the brewnote ui and the help manual. This PR (along with one in the [manual repo](https://github.com/Brewtarget/manual/pull/18)) resolves that inconsistency. I suggest that additional fields and UI changes (as mooted in #54) can be added in future releases with the appropriate DB schema.

I need someone to confirm I've used sane tooltips for Strike Temp and Final Temp.

### DIFF
--- a/ui/brewNoteWidget.ui
+++ b/ui/brewNoteWidget.ui
@@ -211,7 +211,7 @@
                 </size>
                </property>
                <property name="toolTip">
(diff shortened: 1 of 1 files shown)


malavv:
Thanks Kapinga, looks good to me.

I'll look for resolution on #54 before merging.

PR #364, opened 2017-03-23
Closes #94 - Negative Volumes from Mash Designer

### LINKED ISSUE #94: Mash designer - lower step temperature causes negative infusion

kapinga:
Bug #94 notes that the mash designer behaves badly when one step was cooler than the previous step, resulting in negative volumes becoming available. This was due to implicit assumptions within `maxTemp_c()`, `minTemp_c()`, and `minAmt_l()`, all of which assumed that the mash must heat up with each step. This PR fixes this behavior, and adds some enhancements to the Mash Designer besides:

- I've added a heating() private method in MashDesigner that checks if the current step is warmer or cooler than the previous one. The min/max temp/amt functions are all updated to check heating() and then perform the appropriate calculation.
- These functions also now have sanity checks that bound temperatures between 0 C to boiling and volumes between 0 and the amount going into the brew pot.
- These changes opened up additional edge cases, where the user could enter a target temperature that could not be achieved within the bounds above. I created an additional function `maxTargetTemp_c()` that determines the maximum (if heating) or minimum (if cooling) target temperature that can be achieved without overflowing. `saveTargetTemp()` was modified to check the requested target against maxTargetTemp - if the original target is out-of-bounds, it is reset to be the value of maxTargetTemp_c() before the target is saved.
- The typeChanged SLOT call is fixed so that `typeChanged()` actually executes when the user changes the step type.
- I changed default behaviors to be more user friendly:
    - If the user selects a sparge type, the target temp is automatically set to the mash's sparge temp,
      and the time is set to 15 minutes
    - On the first infusion, the amount added is initially set to target a mash thickness of 1.0 qt/lb
    - On all later infusions, the amount added is initially set to the minimum possible
    - On sparges, the amount added is initially set to the maximum possible
- Changes from PR #343 are cherry-picked into this PR, as I made use of the `targetCollectedWortVol_l()` and `targetTotalMashVol_l()` functions.

Things that should probably be checked before this PR gets merged:
1. Someone should do some abuse testing on the Mash Designer to make sure these mods work when the user enters in crazy things (cooling steps, out-of-range targets, etc.).
2. Someone should probably also double check my formula in `maxTargetTemp_c()` to make sure it works right (I derived it from the formula in `tempFromVolume_c()`)
3. Feedback on the new default behaviors is appreciated

### DIFF
--- a/src/MashDesigner.cpp
+++ b/src/MashDesigner.cpp
@@ -309,7 +309,7 @@ double MashDesigner::tempFromVolume_c( double vol_l )
 
 // Returns the maximum possible target temperature that can currently be reached.
 // If cooling, returns the minimum possible target temp
-double MashDesigner::maxTargetTemp()
+double MashDesigner::maxTargetTemp_c()
 {
    if (mashStep == 0 || mash == 0)
       return 0.0;
@@ -624,7 +624,7 @@ void MashDesigner::saveTargetTemp()
       // Do a sanity check the make sure the max volume available can actually hit the temperature we want
       double tempNeeded_c = tempFromVolume_c( maxAmt_l() );
       if ((tempNeeded_c > maxTemp_c()) || (tempNeeded_c < minTemp_c()))
-         lineEdit_temp->setText( maxTargetTemp() );
+         lineEdit_temp->setText( maxTargetTemp_c() );
    }
 
    if ( mashStep != 0 )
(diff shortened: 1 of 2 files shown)


kapinga:
We should also try to pull this into the develop branch when it gets accepted. I'm pretty sure it should merge nicely, but I make no guarantees

kapinga:
I will review my work on this some more and try to make it easier to review when I get some time (no promises on when that will be though)

PR #367, opened 2017-04-30
Added Alcohol Percentage Tool to the Tools menu

rhoob:


dpettersson:
The function of the addition is simple and works. 
My only question is if the change in size in MainWindow.ui.  I am not sure if it matters (i assume that it doesn't). @rocketman768  should get the last word about it.

src/AlcoholTool.cpp
```diff
@@ -0,0 +1,123 @@
+/*
+ * AlcoholTool.cpp is part of Brewtarget and was written by Ryan Hoobler
```
malavv: The copyright text is a bit funky versus what we normally use. How would that be updated when someone else modifies it? I'm not the one who wrote the original statement @rocketman768

src/AlcoholTool.cpp
```diff
@@ -0,0 +1,123 @@
...
+         pushButton_convert = new QPushButton(this);
+            pushButton_convert->setAutoDefault(false);
+            pushButton_convert->setDefault(true);
+         QSpacerItem* verticalSpacer2 = new QSpacerItem(20, 40, QSizePolicy::Minimum, QSizePolicy::Expanding);
+         vLayout->addItem(verticalSpacer);
+         vLayout->addWidget(pushButton_convert);
+         vLayout->addItem(verticalSpacer2);
+
+         QSpacerItem* verticalSpacer3 = new QSpacerItem(20, 40, QSizePolicy::Minimum, QSizePolicy::Expanding);
+         vLayout->addItem(verticalSpacer3);
```
malavv: This layout scheme is very different from what I am used to see and I think different from what this project normally uses. I would suggest using a code formatting tool, but I assume you have done this for readability purposes? I am not sure how other feel but this is harder for me to read.

src/AlcoholTool.cpp
```diff
@@ -0,0 +1,123 @@
...
+{
+    double abv = 0;
+    double og = 0;
+    double fg = 0;
+
+    og = lineEdit_og->toSI();
+    fg = lineEdit_fg->toSI();
+
+    abv =(76.08 * (og-fg) / (1.775-og)) * (fg / 0.794);
+   //http://www.brewersfriend.com/2011/06/16/alcohol-by-volume-calculator-updated/
```
malavv: This conversion of brewers points to predicted ABV is very similar to : 
  `double Algorithms::getABVBySGPlato( double sg, double plato )`

If there is no way to use any of the functions in Algorithms, I would suggest creating it yourself and using it here.

src/AlcoholTool.cpp
malavv: Nevermind, completely wrong. It is used elsewhere, disregard that comment.

src/AlcoholTool.cpp
rocketman768: Unless the author _really_ wants to disown copyright, I agree it should be changed to the standard format:

    * AlcoholTool.cpp is part of Brewtarget, and is Copyright the following
    * authors 2009-2015
    * - Ryan Hoobler <rhoob@yahoo.com>

rocketman768:
Agree with @dpettersson about size changes in mainWindow.ui. They should be reverted unless necessary for some reason.

malavv:
Hi again Rhoob!

I just compiled and tested your new tool, thanks for taking the time to contribute.

From a user standpoint, I might not have the "Reading" part of OG and FG. It might be interesting also to have a placeholder suggesting the format. (ex. 1.014 or 14 depending on what you expect) Also, I now many people who use this and take their OG right at the end of the boil and forget to temperature correct it. It might be interesting to warn people against this.

But the real reason I came to see your Pull Request is that once this is integrated, it would be nice if you could come to Brewtarget/manual and include instruction for your tool in it.

Btw, here is a screenshot of the tool proposed here:
 image

pricelessbrewing:
Giving @rhoob another notification. I believe the PR is still pending due to the copyright issues above, and the UI size change. 

@dpettersson, @rocketman768, if rhoob doesn't come back to edit the PR, what's our move from there? Close PR, reopen issue and have someone else create the tool?

rhoob:
File are changed.  Also added a few words in the manual.

pricelessbrewing:
Awesome, thanks for taking care of this rhoob!

xzfantom:
I think this tool is worth merging.
Should we close this PR and open new one with fixes? Or smbd with write access can fix it in this PR?

matty0ung:
Agree it's worth adding.  Would be nice to tidy it up a bit, eg put the ABV formula in Algorithms, and have it called by Recipe::recalcABV_pct() and AlcoholTool::convert(), rather than implementing the same formula in two different places.

xzfantom:
Yes, this was my thought, too. Keeping all math in one place is good.

mattiasmaahl:
So how do we go about this? The conflicts are not very complicated to resolve and can be done in the web editor.
Do we resolve them, merge and then tidy up with a new PR?
Or do we reach out to Rhoob to make the adjustments and resolving the conflicts?
Suggestions?

matty0ung:
Let's resolve conflicts and merge this fix, and then do a follow-up for the small refactorings.

matty0ung (approved):

matty0ung:
OK, all merged, but needs tidy-up, including fixing field widths on high DPI displays.  Happy for someone else to pick this up, or, if not, I'll try to get to it soonish.

PR #372, opened 2017-07-06
Preparing CMakeList to target QT 5.6

malavv:
Certain modules were previously in use but are no longer necessary or available in QT 5.6.

Their absence means that compiling Brewtarget for 5.6 does not work. This is a short pull to remove the unneeded dependencies.


malavv:
As a check, I built and run the application. It seemed to run fine through some simple manipulation. Then I also looked throught the code base for reference to a webkit object, or to images using the QMngPlugin, but I could not find any.

rocketman768:
@dpettersson what do you think

dpettersson:
Hi,
I am sorry for the late reply. I can take a look at it after work today.

/dpettersson

Den 15 aug. 2017 5:43 fm skrev "Philip Lee" <notifications@github.com>:

dpettersson:
A quick search gave a couple of more references to webkit. Maybe we should change them too before  we merge. I am not sure if it is needed now or if we should create a new case to clean up these and maybe some more old stuff later on. What do you think @rocketman768 ?

It shouldn't be so much job to fix these too.

webkit_search

malavv:
@dpettersson thanks for checking it out.

Looking at the search you did, you seem to have retrieved a lot more results than mine. I think I did mine only on the build folder.

So, I'll fix the Readme and Vagrant, even though they would not be affected by this. It's still a healthy cleanup.

The last 4 results looks a lot like your build folder and some automatically generated docs, both of which will be updated the next time you run them.

malavv:
Small update, I checked the offending **dev-doc** folder, and honestly all of this Windows stuff seems way out of date. For one thing, I think both still provide instructions for QT4. I think any modification to these would do more harm then good.

Maybe, in a later commit, we should completely remove  them and fully commit to having their updated content in the Project's WIKI.

dpettersson:
@malavv I wont argue with you there.
I totally missed that they were in my build-dir.  Thanks for pointing that out. :)

malavv:
@dpettersson do you think that since the changes were applied to the README, this could be merged-in?

PR #374, opened 2017-08-02
fixing BtTextEdit so that it saves text on exit

### LINKED ISSUE #365: Recipe Extras Tab >> Notes & Taste Notes should be saved?
v2.4, Linux mint 18.1
I am assuming that because these sections are editable that they should be saved with the recipe. It's not saving anything I type into these fields.

anderwm:
Not being able to save the notes finally broke me. The textChanged signal was being overridden so it was never triggered, changing the function name works for me. Should fix #365

### DIFF
--- a/src/BtTextEdit.cpp
+++ b/src/BtTextEdit.cpp
@@ -28,7 +28,7 @@ BtTextEdit::BtTextEdit(QWidget *parent)
    wasModified = false;
 
    // We will see if this works...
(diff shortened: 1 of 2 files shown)


dpettersson:
@rocketman768  I have tested the pull request and it looks good.
I suggest we merge.

rocketman768:
@dpettersson thanks! Go ahead and approve/merge.

PR #377, opened 2017-08-19
Removed the Debian /usr/local warning.

malavv:
The CMake code seems to really not like having the CMAKE_INSTALL_PREFIX
as anything else than "/usr/" on Debian but as far as I can see, there
is no reason for it.

Obviously, that must have been done for a reason in 2010, but it's not
obvious to me. Furthermore, the Filesystem Hierarchy Standard
specifically mentions that sys admin should install software in
/usr/local and not /usr, the last being only for read-only files.

Furthermore, I think a newer practice is to install apps in /opt/
instead.

The goal of this push request is mainly to stop CMake shooting at me
every time because I want to install in the standard location.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,8 +99,8 @@ ELSE()
 ENDIF()
 
 IF( APPLE AND NOT CMAKE_OSX_ARCHITECTURES )
-   SET( CMAKE_OSX_ARCHITECTURES x86_64 ) # Build intel 64-bit binary. 
-   #SET( CMAKE_OSX_ARCHITECTURES i386 x86_64 ) # Build intel binary. 
+   SET( CMAKE_OSX_ARCHITECTURES x86_64 ) # Build intel 64-bit binary.
+   #SET( CMAKE_OSX_ARCHITECTURES i386 x86_64 ) # Build intel binary.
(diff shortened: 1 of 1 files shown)


rocketman768:
I think it's fine, minus all the whitespace changes. Revert those, and good to go.

malavv:
Hey @rocketman768 thanks for the really quick review.

The removal of whitespaces at the end of the line is done automatically by the configuration of the `setupgit.sh` at the root of the repository. On line 8, it sets the `core.whitespace` attribute which by default removes EOL whitespaces.

Right now, even if I put back the spaces, Git wouldn't let me commit them. Should I ignore the `setupgit.sh` configuration from now on?

rocketman768:
Ah, I forgot about that. Nevermind then.

dpettersson:
i have looked at the changes and i cannot see any reason why they shouldn't be accepted.
According to https://wiki.debian.org/FilesystemHierarchyStandard /opt is used for  "Add-on application software packages. Pre-compiled, non ".deb" binary distribution (tar'ed..) goes here. "

That indicates that /usr is intended for deb-packages (which is the package format that brewtarget is distributed as under debian and other derived distribution). 

I see no  problem with the check, but i also agree that it is not necessary unless you specifically build for debian. And if you are building a package for debian then you should know how the directory hierarchy works => the warning is not really necessary.

I think that we should merge this. :)

PR #379, opened 2017-09-20
Added temp calibration option to HydrometerTool.cpp

jawsdaws:
My hydrometer is calibrated to 68F.  The hydrometer tool was set to 60F as the calibrated temperature. I added a field to allow the user to edit the calibrated temp.

### DIFF
--- a/src/HydrometerTool.cpp
+++ b/src/HydrometerTool.cpp
@@ -29,11 +29,12 @@
 HydrometerTool::HydrometerTool(QWidget* parent) : QDialog(parent)
 {
    doLayout();
-   
+
    connect( pushButton_convert, &QAbstractButton::clicked, this, &HydrometerTool::convert );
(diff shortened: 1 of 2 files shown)


dpettersson:
I haven't used the hydrometertool before, but it seems to work.
I think we should merge this unless @rocketman768  has any objections.

jawsdaws:
I probably should have put this in the request.  Sorry.  I've not used git or github very much.  
I tested this patch against http://www.straighttothepint.com/hydrometer-temperature-correction/ and https://www.brewersfriend.com/hydrometer-temp/  They almost always gave the same results.  The few times there was a difference, it was 0.001 gravity.  Most likely from rounding or approximating on one side or the other.

rocketman768:
👍

jawsdaws:
Thanks guys! Cheers!

PR #382, opened 2017-11-05
Resolves #94 in develop branch

### LINKED ISSUE #94: Mash designer - lower step temperature causes negative infusion

kapinga:
I got too aggressive in making changes/improvements in my last PR - this is a much more scaled back attempt to resolve #94 in the develop branch.

This PR updates MashDesigner to check if the current step being worked on is heating or cooling the mash. Key changes were then made to `minTemp_c()`, `maxTemp_c()`, and `minAmt_l()` so that these methods behave appropriately when the mash is cooling rather than heating. Sanity checks were updated in a number of other methods - previous assumptions that the mash must be getting hotter were broken.

In addition, I made one small change to update slider positions/values when a new target temperature is entered. This ensures that values held over from the previous step don't carry over if a user does not move either slider on the new step.

All changes are contained to MashDesigner. I have tested it some with some edge cases when designing a mash (create a step that's cooler, a step that's impossibly cool, and a step that's impossibly hot), but someone else should test this out too.

These changes were made on the `develop` branch. Once approved, this update should be back-ported to the `2.4.0` branch to resolve bug #94 in that version too.

### DIFF
--- a/src/MashDesigner.cpp
+++ b/src/MashDesigner.cpp
@@ -600,6 +600,9 @@ void MashDesigner::saveTargetTemp()
    updateMaxAmt();
    updateMinTemp();
    updateMaxTemp();
+   updateAmt();
+   updateTempSlider();
+   updateTemp();
    updateFullness();
    updateCollectedWort();
 }
--- a/src/MashDesigner.cpp
+++ b/src/MashDesigner.cpp
@@ -150,13 +150,8 @@ void MashDesigner::saveStep()
    MashStep::Type type = static_cast<MashStep::Type>(comboBox_type->currentIndex());
    double temp = lineEdit_temp->toSI();
 
-   // TODO: Make this test actually check for the max/min *achievable*
-   // temperature. The max target temp must always be (often much) lower
-   // than the max temperature of water to add...
-   if (heating())
-      temp = std::min(temp, maxTemp_c());
-   else
-      temp = std::max(temp, minTemp_c());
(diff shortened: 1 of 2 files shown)


src/MashDesigner.cpp
```diff
@@ -390,7 +410,7 @@ void MashDesigner::updateFullness()
 double MashDesigner::waterFromMash_l()
 {
    double waterAdded_l = mash->totalMashWater_l();
-   double absorption_lKg;
+   double absorption_lKg, amt;
```
amlynarczyk: This changes nothing

src/MashDesigner.cpp
```diff
@@ -559,10 +574,11 @@ void MashDesigner::updateTemp()
...
 {
    double temp = stepTemp_c();
-   double maxT = maxTemp_c();
 
-   if ( temp > maxT ) 
-      temp = maxT;
+   if (heating())
+      temp = std::min(temp, maxTemp_c());
+   else
+      temp = std::max(temp, minTemp_c());
```
amlynarczyk: This is duplicated, please move to separate method

kapinga:
Good points @amlynarczyk. I reverted the unnecessary change (this was inadvertent), and created an additional private method `double bound_temp_c(double temp_c)` to eliminate the duplicated code.

PR #406, opened 2018-04-21
Use default directory in case of misconfiguration.

MedicMomcilo:
If brewtarget directory is configured but non-existent, we should fall back to default location.
Use case for this is when brewtarget directory is shared amongst many users.
Since config file takes absolute path, we can't just set ~/.config/brewtarget to accommodate.

With this patch if directory is misconfigured or doesn't exist, we would use the same guessing mechanism.

Please let me know if there is anything I still need to do and thank you for this awesome piece of software.

### DIFF
--- a/src/brewtarget.cpp
+++ b/src/brewtarget.cpp
@@ -404,11 +404,12 @@ bool Brewtarget::initialize(const QString &userDirectory)
       userDataDir = QDir(userDirectory).canonicalPath();
    }
    // Use directory from app settings.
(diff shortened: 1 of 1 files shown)


malavv:
The title of your PR is a bit misleading with regard to the modifications you have made. I was worried your modifications might impact new developers misconfiguring test environments and using their main database without knowing.

Nonetheless, I think only using the saved configuration if it is a valid directory is the expected behaviour., at least for me.

lgtm

PR #413, opened 2018-07-08
Improve fermentable, hop and yeast editor layout

peterurbanec:
There are several fields in these editors that do not grow when the window
is resized larger or are not big enough to fit content. Increase the max
size of these fields so that they can accommodate the content.


peterurbanec:
The Travis CI build error appears unrelated to the changes in this pull request.

dpettersson:
I agree. The build error seems to be related to QT and not this change.

Issue #340, opened 2017-01-19
BtTreeView selected items take precendence over locally selected entities


eltomek:
Found on 2.4.0
I had some problems trying to figure out the headline, here are the steps:
1. open some recipe to be edited and switch to `Recipe` tab of the main window
2. in the left hand-side `BtTreeView` tree select the Equipment tab and double-click some entry that is different to the equipment selected in the Equipment field of the recipe
3. an Equipment edit dialog will pop up, press Cancel
4. click on the currently selected Equipment button of the recipe to open the Equipment edit dialog (**issue** is that the Equipment edit dialog now calls for the one that you double-clicked in the  `BtTreeView`, not the one that you had picked for your recipe. The same works for style.

pricelessbrewing:
Confirmed still an issue.

PR #424, opened 2018-09-22
Resolving bug 340: The current equipment needed to be set before the …

dpettersson:
The editor retained the latest shown equipment. It seems that all it took was  to set the current recipe equipment in the editor before showing it.


dpettersson:
Closes #340

dpettersson:
Closes #340: The current equipment needed to be set before the …

PR #429, opened 2018-10-02
Closes #158. Imported the patch from develop

### LINKED ISSUE #158: Initial value of IBU Adjustments erronous on system use coma as decimal separator

dpettersson:
Imported the patch from develop.
It looked ok on my computer.

@pricelessbrewing review and merge?

### DIFF
--- a/src/BtLineEdit.cpp
+++ b/src/BtLineEdit.cpp
@@ -41,7 +41,7 @@ BtLineEdit::BtLineEdit(QWidget *parent, Unit::UnitType type) :
    _forceUnit = Unit::noUnit;
    _forceScale = Unit::noScale;
    */
-   
+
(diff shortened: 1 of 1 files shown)


pricelessbrewing:
Are we keeping 2.4 branch even though it's not officially released yet and we still have pending issues on dev branch?

dpettersson:
I imported it since you had the case in the 2.4 project,
It shouldn't cause any problems when we merge 2.4 back to main after
release.

Den ons 3 okt. 2018 kl 04:39 skrev Mark Price <notifications@github.com>:

PR #447, opened 2019-07-26
missing flag

idarlund:
Added missing flag: Norway.
Got from https://commons.wikimedia.org/wiki/File:Flag_of_Norway.svg
License: This image of simple geometry is ineligible for copyright and therefore in the public domain, because it consists entirely of information that is common property and contains no original authorship.


idarlund:
this won't work. need to change size and format.

PR #453, opened 2019-12-12
Cache

mikfire:
Merry Christmas/Saturnalis/Festivus! I come bearing a massive pull request that I've been working on for about a year now.

This pull request does 5 main things:
   1. All BeerXMLElements cache their properties. Reads pull from memory instead of the database;
   2. All of the maps, lists, vectors, etc. that defined database tables are gone and replaced with three new classes;
   3. The inventory system is completely reworked to deal play nice with things 1 & 2;
   4. DatabaseSchemaHelper has been reduced to a shell of its old self;
   5. Several small fixes to make our code a little more modern

The benefits of this are:
   1. My startup times using a local postgresql db have gone from 10 seconds to somewhere between 1 and 2 seconds;
   2. Adding a new column to a table now only requires you to edit 3 files, and they are well defined edits. Previously, you needed to edit 7 - 8 files in a lot of places to get it right.

Be warned. This will bump the database version. It will drop a table, drop a few columns and completely re-organize the inventory tables. I *strongly* recommend testing this only with a safe backup.

If you want to understand more, there is a new file in dev-docs called database.markdown that explains why I've done what I did.


mikfire:
No takers?

dpettersson:
Hi,

I can test it tonight when i get home from work.

I must say i do feel bad for not doing anything for quite a while.
We really should get the new release out sion.

/Daniel P

Den tis 28 jan. 2020 03:15mikfire <notifications@github.com> skrev:

dpettersson:
Hi again.
This a really massive pull request. :)

I downloaded it, patched and tried to test it locally. All i was met by was a prompt for the password for the database. Does this require a real dbms or can we still use the sqlite-files?

I will continue looking at the code, but as you might expect it might take a day or two.

src/BeerXMLElement.h
```diff
@@ -57,7 +58,7 @@ class BeerXMLElement : public QObject
 
    friend class Database;
 public:
-   BeerXMLElement(Brewtarget::DBTable table, int key);
+   BeerXMLElement(Brewtarget::DBTable table, int key, QString t_name = QString(), bool t_display = false, QString folder = QString());
```
dpettersson: The line is really long. 
May i be an annoying bastard and ask you to split it into several lines?
It is not required but it would probably help with the readability

mikfire: I will work on the formatting.

If you are getting prompted for a password, I will promise your dbType is 2 (postgres) right now, and it sounds like you want it to be 1 (sqlite). That was one of the more uncomfortable changes I made, and it seems it didn't work as expected.

The quick fix will be to add config_version=1 and set dbType=1 in the [General] section of the config. Obviously, I am not suggesting we ship the code like this. I will fix it and figure out what broke.

mikfire:
I found and fixed the dbType problem. Your best bet is to either manually set it to 1. Alternately, you can remove both it and the config_version key/value pairs from the config file and try again.

dpettersson:
Sorry for the delay. I promise the turn around will be faster in the future (partly because i will be getting fridays off at least until summer).
Unfortunately i still  have a problem when i test your changes.

As you know i am still using the "old style of database". 
I tested adding a recipe and when i click OK in the popup that asks for a name for the brew i get the following:
[10:49:32.249] ERROR : int Database::insertElement(BeerXMLElement*) could not insert a record into recipe: INSERT INTO recipe (age,age_temp,assistant_brewer,batch_size,boil_size,boil_time,brewer,carbonationtemp_c,carb_volume,date,deleted,display,efficiency,fermentation_stages,fg,folder,forced_carb,keg_priming_factor,locked,name,notes,og,primary_age,primary_temp,priming_sugar_equiv,priming_sugar_name,secondary_age,secondary_temp,taste_notes,taste_rating,tertiary_age,tertiary_temp,type) VALUES(:age,:ageTemp_c,:asstBrewer,:batchSize_l,:boilSize_l,:boilTime_min,:brewer,:carbonationTemp_c,:carbonation_vols,:date,:deleted,:display,:efficiency_pct,:fermentationStages,:fg,:folder,:forcedCarb,:kegPrimingFactor,:locked,:name,:notes,:og,:primaryAge_days,:primaryTemp_c,:primingSugarEquiv,:primingSugarName,:secondaryAge_days,:secondaryTemp_c,:tasteNotes,:tasteRating,:tertiaryAge_days,:tertiaryTemp_c,:type)  **Parameter count mismatch**
 
I have to say that the startup is getting much faster than i can remember it was when you started improving the database connectivity.

mikfire:
It does :) but the code should take care of all of that for you. Check to see the version number in the settings table. I think it should be something like 8 after the code runs, mostly to get the inventory stuff worked around. I think, otherwise, it just removes a few unused columns from brewnotes.

Can you open up your sqlite db, look at the recipe table and see if it has a column named 'locked'? I cannot reproduce, but I have a sneaking suspicion this is the issue.

dpettersson:
That's better :)
I have now tried creating a recipe, brewing and then deleting. Everything works :)

PR #458, opened 2020-04-08
Bugs/bnotes text

### LINKED ISSUE #457: Text Notes on Brew Days gone and do not work anymore
Before #453 I was able to store notes against each brew day.

After upgrading all the notes have gone blank. If I attempt to save a note, then an error is thrown and the application crashes:
image

```
[17:22:55.327] ERROR : Could not translate notes to a column name
Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
If that is not possible, in Qt 5 you must at least reimplement
QCoreApplication::notify() and catch all exceptions there.
```

Steps to reproduce:

1. click on a recipe
2. click "Brew It" or select an existing brew day
3. click on the relevant date tab, then enter any text in `Notes` box
4. click away from the notes box

The schema of my brewnotes table is this:

```
CREATE TABLE "brewnote" 
(
    id INTEGER PRIMARY KEY autoincrement ,
    abv real DEFAULT 0,
    attenuation real DEFAULT 1,
    boil_off real DEFAULT 1,
    brewdate timestamp DEFAULT CURRENT_TIMESTAMP,
    brewhouse_eff real DEFAULT 70,
    deleted boolean DEFAULT 0,
    display boolean DEFAULT 1,
    eff_into_bk real DEFAULT 70,
    fermentdate timestamp DEFAULT CURRENT_TIMESTAMP,
    fg real DEFAULT 1,
    final_volume real DEFAULT 1,
    folder text DEFAULT '',
    mash_final_temp real DEFAULT 67,
    og real DEFAULT 1,
    pitch_temp real DEFAULT 20,
    post_boil_volume real DEFAULT 0,
    projected_abv real DEFAULT 1,
    projected_atten real DEFAULT 75,
    projected_boil_grav real DEFAULT 1,
    projected_eff real DEFAULT 1,
    projected_ferm_points real DEFAULT 1,
    projected_fg real DEFAULT 1,
    projected_mash_fin_temp real DEFAULT 67,
    projected_og real DEFAULT 1,
    projected_points real DEFAULT 1,
    projected_strike_temp real DEFAULT 70,
    projected_vol_into_bk real DEFAULT 1,
    projected_vol_into_ferm real DEFAULT 0,
    sg real DEFAULT 1,
    strike_temp real DEFAULT 70,
    volume_into_bk real DEFAULT 0,
    volume_into_fermenter real DEFAULT 0,
    recipe_id integer,
    FOREIGN KEY(recipe_id) REFERENCES recipe(id)
)
```

If I add a `TEXT` field called `notes` to the `brewnote` table, then the contents of this field will show up in the Notes box, but if I attempt to add any of it or edit it, then the error above happens and Brewtarget crashes.

Notes on recipes still appear to work.

mikfire:
Closes #457.


dpettersson:
Unfortunately i think that this might need some more work.
I downloaded the master and applied the fix.
It started ok, but when i entered a brenote on a recipe:
image

cgspeck:
If I rollback `database.sqlite` to schema version `7` (pre #453 ) then when I run this:

* my existing brewday notes appear (yay!);
* I am able to edit them;
* I am able to create new ones;
* no crashes; and
* the database schema version is upgraded to `8`.

If I use an existing version `8` database (e.g. running unpatched `devel`), then run a version of Brewtarget with this patch, I get the same error as @dpettersson  [above](https://github.com/Brewtarget/brewtarget/pull/458#issuecomment-611927678). I assume this is because the `brewnote.notes` column is missing.

It looks like there might need to be a schema migration to add the notes column back if it doesn't exist and perhaps bump the schema version to 9.

mikfire:
You are correct -- the previously upgraded version will have the missing notes table. My fix did not put the missing column back in. It only fixed the table definition, which fixed the upgrade from 7 -> 8.

If you are willing to screw around a little. Using the "broken" `8` database, manually set settings.version to `7` and then run brewtarget again. I am honestly uncertain what will happen, but it might just fix the problem

Since this is only in the dev branch for right now, I am not sure we need a fix? My next pull request, if ever I get it done, will bump the version to 9. I can put a check into that upgrade to re-add the column if it is missing. Would that work?

cgspeck:
I attempted setting the version of a "broken" `8` db to `7`, then starting the patched version of Brewtarget but it crashed with this in the log:

```
[14:45:38.356] INFO : rearranging inventory
[14:45:38.356] INFO : dropping inventory columns
[14:45:38.356] INFO : dropping bt_alltables
[14:45:38.356] ERROR : Rolling back
[14:45:38.356] ERROR : Database migration 7->8 failed
[14:46:19.947] INFO : Brewtarget is cleaning up.
brewtarget_log.txt (END)
```

I can workaround this by rolling back to my real version 7 database (which gets upgraded to "fixed" version 8). Alternatively I guess one could manually add the column to a broken version 8 db.

I think this is a question for @dpettersson as to whether these workaround are sufficient and whether the column check can be deferred to the later PR.

PR #468, opened 2020-07-30
Chemistry

mikfire:
So. Yeah. Who wanted a builtin water spreadsheet?

I am fairly comfortable with the maths. I've tested against Kai's spreadsheet and I seem to be accurate enough. pH can be a little twisty, but I am willing to ignore that as round off errors.

There is no link between this work and the misc table. There should be, but I wasted 6months trying to figure some way of doing that and all my solutions sucked. This works, and I'm shipping as MVP.


dpettersson:
Testing :)
Unfortunately i might have found some strange stuff. 
I will update this in a couple of hours.

mikfire:
The calculator makes pretty heavy use of the selected recipe's mash settings. If you haven't defined a mash for the recipe, things are really not going to work.

I will make the warnings more obvious, and I also need to consider what happens when brewtarget starts and there is no selected recipe.

mikfire:
And I've just realized I have other problems with no sparge. So I will work on that too

dpettersson:
It looks like we have a problem. Could you take a look at it? 
It is the inclusion of 2 header files.

PR #479, opened 2020-09-11
closes #474 Adding logging option

### LINKED ISSUE #474: Feature request: Add logging options to GUI
Is there an interest in having logging settings in the GUI for the user?

My thinking here is that if we add options to turn on logging, i.e. INFO, WARNING, ERROR and DEBUG, in the GUI and letting the user choose where to put the logs, we can have more end-user bug reporting with logs to support the developer in troubleshooting.

I was tinkering a little on how that would look below:
bild

then adding checks to the logging functions to see if the user wants to log and to what level?
for example:
brewtarget.cpp:
```cpp
void Brewtarget::logE( QString message )
{
   if ( Log::LoggingEnabled && Log:LogLevel <= Log::LoggingLevels.ERROR) 
   {
      log.error(message);
   }
}
```

I began to code for it as I found it to be useful, but what do you guys think?
Is it worth giving the end-user the ability to submit bug reports including detailed logs?

mattiasmaahl:
I've done some coding towards adding functionality to the GUI to let the user decide logging level and where to put the log files.
I mentioned this in #474 asking if it was of interest.

So I created this PR, having no expectation that it will be merged.

I also had a thought that we could implement a command-line arguments, i.e. brewtarget --logging, to activate the logging options for debugging maybe?

All feedback is most welcome!

closes #474

### DIFF
--- a/src/Log.cpp
+++ b/src/Log.cpp
@@ -64,6 +64,43 @@ void Log::changeDirectory(const QDir defaultDir) {
    warn(QString("Could not create a log file."));
 }
 
+void Log::changeDirectory() {
+   //If it's the same, just return, no need to do anything.
+   if ( LogFilePath.filePath(filename) == file.fileName() ) {
+      return;
+   }
+
+   //If the file is already initialize, it needs to be closed and redefined.
+   if (stream) {
+      delete stream;
+      stream = nullptr;
+      if( file.isOpen() )
+         file.close();
+      //Preserving the old logfile
+      if ( ! file.copy(LogFilePath.filePath(filename)) ) {
+         error("Error while copying to the new file location\nReverting settings");
+         LogFilePath = QDir(QFileInfo(file).filePath());
+      }
+   }
+
+   // Test default location
+   file.setFileName(LogFilePath.filePath(filename));
+   if( file.open(QIODevice::Append) ) {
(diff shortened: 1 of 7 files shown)


mikfire:
I am looking at this, and need to poke at the code a little. Do you feel this is ready to commit, or is this simply to drive comments?

mattiasmaahl:
You are most welcome to poke the code as much as you want. I put it up mostly for comments and addition.
I'll take all the feedback/help I can! 😊

mattiasmaahl:
I brought the branch up-to-date with develop today.

As for if I feel if it's ready for commit, yes as is it builds and works as I intended, but there are some thoughts I wanted to run by you folks first.
1. Do you want the option visible at all times or hide it behind a command-line option.
2. I have not tested the code on Mac or *nix, but it should work.
3. do you want the functionality in the first place.

Note on 3. As I wrote earlier, I wrote the code for this as I find it useful to have logs when troubleshooting bugs, so giving users the option to enable logging at run-time and having that when they submit issues/bug-reports would be helpful to us as developers.

For future development I was thinking, and this is only an idea, of maybe having a "submit bug report" button as well within the software making it easier for users to leave them.

mikfire (approved):

PR #480, opened 2020-09-11
Splits XML code from Database, with extras

mikfire:
database.cpp was about 5200 lines prior to this, excluding the header file and it's template methods. I have been thinking about how to reduce the size for a while, and the best slice was to isolate the XML into it's own class. That cleaned out about 1600 lines.

beerJson is coming. Adding more lines to database.cpp wasn't making me happy. This way, we can create a beerJSON class just like beerXML. 

I also renamed BeerXMLElement to Ingredient. I wanted the BeerXML name for my new class, and I have long thought BeerXMLElement needed renaming. So I did.

I also had to fix some problems from the last database refactor I did. The XML wasn't "properly" outputting the type, form and use fields for any object. That has been fixed.

**IMPORTANT**: Check your database and make sure the style table still has the category_number column. It wasn't defined in TableSchema::defineStyle() method and it was missing from my databases. It is still defined in the default, so I am hoping this was just my problem. You can run this via sqlite3:
```SQL
select id,name,category_number from style where name='American Pale Ale'; 
```
and they should all have 10 as the category_number. If not, PM me and I can give you a quick fix. 

Part of this commit fixes that issue. If desired, I can make that a separate PR.


mattiasmaahl:
As I read through this I cannot find the files:
* BeerXML.h/cpp
* Ingredient.h/cpp
* IngredientSortProxyModel.h/cpp

Seems the newly created/renamed files are missing? Or am I missing something?
Everything else is looking good! :)

mikfire:
Yes. You missed the part where I'm an idiot and forgot to git add the new files.

mattiasmaahl:
No worries mate! I'll go through the code tomorrow and review it for you! 😊

mattiasmaahl (approved):
Looks good to me, some Copyright text where your name is not included but other than that it's good.

PR #487, opened 2020-09-17
Fixes: #486 - DataDirectory now pointing to Appdata\roaming on windows clients.

### LINKED ISSUE #486: Bug With database not creating in windows on fresh install
I found that the logfile and database was not able to be created when Brewtarget was installed on a fresh computer where BT had not been installed before.

I'm in the process of fixing a PR for that bug. will be up soon.

mattiasmaahl:
Database directory pointing to the wrong location on Windows clients.

On new installation in windows, where there was no old db to convert, brewtarget was trying to create the new DB in the Programs directory. That directory, which for most normal users, is a write-restricted directory.
I changed the default location to the Appdata\roaming directory on windows builds instead while keeping the original data direktory intact in Linux and MacOS.

This issue is only showing up when you install brewtarget to the Default location C:\Program Files (86)\ or C:\Program Files\

this PR fixes issue #486 
Closes #486

### DIFF
--- a/cmake/modules/NSIS.template.in
+++ b/cmake/modules/NSIS.template.in
@@ -435,6 +435,42 @@ FunctionEnd
 !insertmacro IsNT ""
 !insertmacro IsNT "un."
 
+
+
+; ConvertToCanonicalPath
+; input, top of stack = Path string to convert to Canonical path
+; usage:
+;     Push "C:\Path\to\Folder"
+;     Call ConvertToCanonicalPath
+;     Pop $R0
+;     $R0 will be C:/Path/to/Folder
+
+Function ConvertToCanonicalPath
+   Exch $R0 ;input string
+   Push $R1
+   ${StrRep} $R1 $R0 "\" "/"
+   StrCpy $R0 $R1
+   Pop $R1
+   Exch $R0 ;output string
+FunctionEnd
+
+; ConvertFromCanonicalPath
+; input, top of stack = Path string to convert to Canonical path
+; usage:
+;     Push "C:/Path/to/Folder"
+;     Call ConvertFromCanonicalPath
+;     Pop $R0
+;     $R0 will be C:\Path\to\Folder
+
+Function ConvertFromCanonicalPath
+   Exch $R0 ;input string
+   Push $R1
+   ${StrRep} $R1 $R0 "/" "\"
+   StrCpy $R0 $R1
+   Pop $R1
+   Exch $R0 ;output string
+FunctionEnd
+
 ; StrStr
 ; input, top of stack = string to search for
 ;        top of stack-1 = string to search in
@@ -752,6 +788,7 @@ FunctionEnd
 ; brewtarget specific instructions
 
 Var btPath
+Var btWindowsPath
 Var btPathVersion
 var prevInst
 Var targetDir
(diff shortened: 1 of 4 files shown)


mikfire:
Is there a way to make this right on the upgrade as well? Perhaps copy from the old location to the new via the NSIS installer?

mattiasmaahl:
I could look into that, haven't look at it at all actually.
will revert when I have a better answer to you.

mattiasmaahl:
As I read the NSIS.template.in file the function is already doing what you suggest.
1. Look for old installation, both in registry and user supplied path if not found in Registry.
2. copy the old files from Local Appdata or user supplied path to Roaming Appdata
3. sets the flag "hadOldConfig" to the registry for the application to upgrade the database upon first startup.

Was that what you were asking for or did I misunderstand you?

Code can be found in cmake/modules/NSIS.Templatefile between rows 751-961

Question is, if we also should poke the new "user-data-dir" into the registry because I could not find that this was done at any point during the copying of files, and on brewtarget startup it will check the registry before defaulting to Appdata Roaming. that will in the case upgrading from 2.3.x look at the old user-data-dir instead of the new one.
Haven't tested this yet though.

mattiasmaahl:
And I'm also blind, the value for the new user-data-dir is written to the registry in the OrganizationDefault on row 922 in NSIS.template.in file, although is that really where brewtarget will check for settings? 

My understanding is that all settings read from the registry is read from the HKCU\SOFTWARE\brewtarget path? or am I missing something?

mattiasmaahl:
After checking closer on this, I cannot find any referens in the QSettings dokumentation to the "OrganizationDefaults" that QSettings will check that path for any reason, which leaves me to think that those values written there will never be used. 
Nowhere in the code is there a reference to the OrganizationDefaults either.

Either we rewrite some of the Settings code in the source or we rewrite the NSIS template to write the keys to the registry in the right place. any thoughts?

mattiasmaahl:
So, On a row in the documentation I found that OrganizationDefaults is a thing, so reverted the change that did that modification.

The NSIS.template.in file should be OK with this PR as is. It should copy old Databases to the new location.
I will run a few tests to verify that it works as I read it also, but that will have to wait until next week though.

mattiasmaahl:
Sorry for the Confusing reverts, I reverted the wrong commit the first time!
**embarrising**

mattiasmaahl:
Ok, so after some digging in the system processes I found that Brewtarget-2.3.0 on Windows stores the database in 
Users-Appdata\local\VirtualStore\Program files (x86)\Brewtarget-2.3.0\data directory. ( <- Windows-magix )

That directory is not found by the NSIS installer, will try to fix that. also, when I point the NSIS installer to the new location everything is copied into the new location and OrganizationDefaults is updated as intended, BUT in the registry the userdatadir key in brewtarget settings is still pointing to the old location (c:\program files (x86)\brewtarget-2.3.0\data) which BTW is not the location of the database! there's some windows magic going on here in the background putting the database in the VirtualStore directory.

will try to fix the NSIS installer to look into the VirtualStore (if exist) and also erase (or rewrite) the old reg-key from brewtarget registry settings.

will put up a commit for that and also try to clean up my brain-fart from last week! =)

mikfire:
If you could do that, I would be endlessly grateful and will buy you a pint should I ever make it to your part of the world. I struggled to figure out how that worked a while ago, and failed. The windows magix is really obscure to me.

If we can get that working better, I will backport it to the 240 release and possibly roll it.

mattiasmaahl:
This last commit should fix the NSIS installer, The issue eluded me for a day or two :) but I added a Section that checks the $prevInst flag before ever trying to start the Directorypage for user input on where the datapath is.

then setting the $prevInst flag to "CopyDone" prevents that last dialog from appearing.

One note on this is that Brewtarget stores the paths in Canonical format, i.e. C:/Users/.... and all the comparisons against the systems environment variables fails because C:/Program Files (x86)/ IS NOT equal to C:\Program Files (x86)
so I created two functions that will convert from and to Canonical format on the paths.

This also had a small side-effect that the tasks carried out by the btCopyAllFiles is shown in the installations-dialog

I have tested this with upgrade from 2.3.0. do we need to test against the earlier versions as well? 
or what do you think @mikfire ?

mikfire:
I am good with testing from 2.3.0 (or 2.3.1) only. I do not recall any changes to the installer prior to 240. If it works from 230->240, I would expect it to work from earlier.

mattiasmaahl:
I believe so as well, I kept all the checks for prior versions and making sure the earlier XML-files is moved.

mikfire:
It looks good. Give me a bit to test a few things, and I will approve and merge.

mikfire (approved):

PR #491, opened 2020-10-01
Fixing Windows 7 build in Develop-branch

mattiasmaahl:
This PR will fix the windows build to work on windows 7+

added the vcredist_x64.exe and also modified NSIS to include and install automatically

### DIFF
--- a/cmake/modules/NSIS.template.in
+++ b/cmake/modules/NSIS.template.in
@@ -6,6 +6,7 @@
   !define VERSION "@CPACK_PACKAGE_VERSION@"
   !define PATCH  "@CPACK_PACKAGE_VERSION_PATCH@"
   !define INST_DIR "@CPACK_TEMPORARY_DIRECTORY@"
+  !define REDIST_DIR "@CPACK_PACKAGE_REDIST_DIR@"
 
 ;--------------------------------
 ;Variables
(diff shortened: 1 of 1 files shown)


xzfantom:
As far as I understand, after  compiling BT in MinGW there is no need in vcredist. Do you build project in Visual Studio?

mattiasmaahl:
@xzfantom, Yes, I do build BT in V.Studio for the windows build. I will test your way and try to run the application on a fresh win7 to see if you're correct in not needing the DLLs from the redist package.
One question though, do you Cross-compile or MinGW under windows?
Also, I'm compiling under Windows 10, I assume you do as well?

xzfantom:
@mattiasmaahl I do both way - on laptop with Win10 I use VSC + MinGW, on linuxbox (kind of multimedia server with Ubuntu 20.04) I just experiment with building binaries cross-compile with help of MXE (no actual development there). In addition I try to cross-compile in GitHub Actions - like here https://github.com/xzfantom/brewtarget/actions/runs/513245846 (but smth happened to deb build there, win build works fine). GitHub provide windows box with VS, so it's possible to build binaries with help of VS, but I don't know how) C++ is not my main language, so

mattiasmaahl:
Ah, OK. well when I'm building in VS I am using MSVC_x64 to build the binaries, maybe the issue with the DLLs reside therein.
I will test out building with MinGW instead and revert.
It may be the case that the MSVC compiler is only dynamically linking the DLLs from the system making it an requirement to have the redist package installed.

mattiasmaahl:
@mikfire this is not valid anymore, let's close it and move on with our lives. 😊

PR #492, opened 2020-10-25
Fix for Issue #497 – (Bugs 1109751 and 1372087 from launchpad)

matty0ung:
(This is my first pull request for brewtarget, so grateful for any feedback about process, standards, etc.)
Bug #1109751 is basically a small feature request for the user to be able to choose the file name into which a manual DB backup is created.
I modified database class to:
 - Be able to report the default name it uses for the backup file - new method getDefaultBackupFileName()
 - Be able to do a backup to a fully qualified file name - new method backupToFile()
 - Have existing method backupToDir() work by calling backupToFile() - to avoid duplicate code
 - Add some debug-level logging
I modified MainWindow::backup() to:
 - Ask the user to select a file instead of a directory, but have the default filename already in the dialog - so the user doesn't have to specify anything more than now if they are happy with the default
 - Correct what happens if the user clicks 'Cancel' on the dialog.  (Currently, the user sees an error message, but it seems better for this to just cancel the backup.)
 - Add some debug-level logging

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1894,12 +1894,19 @@ void MainWindow::brewAgainHelper()
 
 void MainWindow::backup()
 {
-   QString dir = QFileDialog::getExistingDirectory(this, tr("Backup Database"));
+   // NB: QDir does all the necessary magic of translating '/' to whatever current platform's directory separator is
+   QString defaultBackupFileName = QDir::currentPath() + "/" + Database::getDefaultBackupFileName();
+   QString backupFileName = QFileDialog::getSaveFileName(this, tr("Backup Database"), defaultBackupFileName);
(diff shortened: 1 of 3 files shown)


matty0ung:
I see now that there seem to be two different places that Brewtarget bugs are logged - https://bugs.launchpad.net/brewtarget and https://github.com/Brewtarget/brewtarget/issues.  Are both valid or is one old?

dpettersson:
Hi,

I have always used the one at github.
I cant remember if i has looked at the other list.

/Daniel P

Den mån 26 okt. 2020 08:49matty0ung <notifications@github.com> skrev:

matty0ung:
Ah, thanks, good to know.  I'll make a ticket about the Launchpad one.

mikfire (approved):

Issue #370, opened 2017-05-24
Request: Undo/Redo Ability


marker5a:
As a user, when formulating recipes, it can be easy to to accidentally make a change to a recipe unintentionally.  At the moment, there is no good way, that I can see, to undo or redo the change.  When I make a mistake to a recipe, I generally use the clunkier approach of just restarting the program and risk losing unsaved changes to the recipe.

It seems that it might be nice to track changes made in a stack (or stacks for redo and undo) datatype and store "transactions" in these LIFO buffers.  This is a potentially larger task as it would require user actions to be distilled into a transactional datatype to be placed in the LIFO.

Any ideas or suggestions?  I'd like to work on this task, but would like to get some feedback from users as to the feasibility and/or how useful this would even be to other users.

speshak:
I'm not sure how feasible it is to use, but QT5 does have an undo framework that might be worth looking at: http://doc.qt.io/qt-5/qundo.html

marker5a:
I like that idea of not having to build a new datatype... would be much nicer to use the built in framework for the heavy lifting.  I need to dig into it further to understand the scope of what it can do, but it sounds promising!

In any case, I'm still going through the code to find a common way to handle different transactions.  For instance, removing an item from the Fermentables table is a different type of operation than changing the name of a recipe.  One thought I have is to fill up the stack (or QUndo) with callback functions to undo the task.  Along with this would be a set of argument for the callback to perform it's function.

For instance, deleting a row in the fermentables table would push to the stack a callback for adding a fermentable with the arguments that would be needed to complete that transaction identically.  Same would go for if someone renamed a recipe... you would just push a callback pointer to changing the recipe name and it's argument would be the old recipe name.

matty0ung:
FYI I am working on an implementation of this, using the QT5 undo framework as suggested above.  I have simple cases working (edit a value).  Adding/removing fermentables etc will be a bit more fun, but should be doable.

matty0ung:
Making progress on this.  I have undo/redo pretty much working for adding/removing ingredients, which turned out to require touching a lot of different places in the codebase.  Need to check a few edge cases and tidy up a couple of things but hope to have a pull request in the not too distant future.

mikfire:
Just out of curiosity, have you tested this against postgres as well? Since I am the only person who likely uses it, "No" is a perfectly acceptable answer.

matty0ung:
[...]

PR #499, opened 2020-11-24
Feature 370 - Add Undo/Redo capability

matty0ung:
This is a first-pass implementation of Feature https://github.com/Brewtarget/brewtarget/issues/370.  It makes a lot of things undoable (and redoable), but leaves room for further enhancements in future.  (I didn't want to make this pull request any larger!)

We're using the Qt undo framework (see https://doc.qt.io/qt-5/qundo.html).  This saves us reinventing the wheel, but we need to make a fair number of small changes to fit with the model of that framework.   Essentially, to make an action (eg edit recipe name) undoable (and redoable), you have to turn it into a QUndoCommand object and push it onto QUndoStack.  Pushing the command on the stack will cause it to be executed (Do).  You can then ask the stack to Undo (or Redo) the last (or next) command.

So, aside from adding a QUndoStack to MainWindow, and doing a bit of plumbing to make the new menu entries (Edit > Undo / Edit > Redo) behave properly (eg grayed-out when not available etc, plus show the name of the relevant command they would do), all we needed to do was, for every action (eg Add Fermentable to Recipe) replace the code that does that action with code that create an object of some subclass of QUndoCommand.  Simple - except that every QUndoCommand object you create needs to know how to undo itself.  (Basically, in the QUndoCommand object you implement Do (= Redo) and Undo.)

The easy stuff is handled by creating instances of new class SimpleUndoableUpdate. This can handle any action that boils down to changing a Q_PROPERTY - eg changing recipe boil time or brewer name. Fortunately this is a lot of actions.  All the class has to do is be passed an object (eg the recipe), a parameter name (eg brewer), and the value the parameter is being set to.  Then with the magic of Qt metaObjects we can call the getter for that parameter (to retrieve the current value) and then (Re)do / Undo are just 'set new value' and 'set old value'.

In some cases, what we're setting is a pointer to an object, which doesn't seem to be something that Q_PROPERTY can handle.  So, for these cases, another new class, RelationalUndoableUpdate, uses templates to do pretty much the same thing as SimpleUndoableUpdate at the expense of the caller having to pass more information in (pointer to setter member function etc).  Also, sometimes doing such an update has extra GUI consequences, so we need a callback to MainWindow after (Re)do or Undo.  Best example of this is when you change recipe style and the OG/FG/etc bar ranges update accordingly on the display.

The final case, which is a bit more complicated, is adding to, or removing from, a list of attributes -- eg adding a fermentable to a recipe.  New template class UndoableAddOrRemove handles these sorts of changes in a similar way to RelationalUndoableUpdate.  (Instead of storing an old and a new value, we have one value and we’re either calling an adder or a remover.  The callback, if any, to MainWindow depends on whether we were doing or undoing and add or remove.

So, the new QundoCommand subclasses are not a huge amount of code.  However, I’ve had to touch a lot of other places in the codebase to tweak edit/add/remove actions into a sufficiently common format that they can plug-in to such QundoCommands.  I was trying to hold to the principle that the undo/redo framework should not really have to know the inner workings of database/UI/main object model.  Without making change for the sake of it, when I’ve had to modify things, it’s generally been in the direction of having the UI code interact with the main object model (Recipe, Fermentable, Mash, etc) and leave that object model to ask Database class to poke around in the database.  (In many cases this is already what happens, but a few times I’ve pushed calls to the Database class inside objects.)

Will do more detailed comments in a subsequent post on this pull request!


matty0ung:
As promised, here are some more details:

This is a non-trivial pull-request, but it looks bigger than it is for a couple of reasons.

Firstly, a lot of the changes are, effectively, the same change applied in multiple places. In a lot of places in the GUI, eg FermentableDialog::addFermentable(), we replace a direct call to a Database class member function (addToRecipe) with a call to a MainWindow member function.  This is because it is MainWindow that has the undo/redo controls and needs to keep the Undo/Redo menus in the correct state.

Secondly, the changes to database.h and database.cpp are not as large as they appear.  I simply moved the definitions of some (mostly templated) functions from database.h to database.cpp – specifically those that are only invoked in database.cpp:
   template <class T> void Database::populateElements()
   template <class T> bool Database::getElements()
   QMetaProperty Database::metaProperty()
   template<class T> T* Database::addIngredientToRecipe()
   QVariant Database::get()
   template<class T> T* Database::copy()
With the exception of adding some comments to the last one, I haven’t changed the implementation of these functions.  The database.h header is now somewhat smaller.

Also, in various places I added extra debug logs, including to help me understand program flow. Happy to comment out some of these if you think it makes the debug logs too verbose.

There are some substantive changes though.  To fit with the Qt Undo Framework, I have made some changes to how ‘add’ and ‘remove’ functions work, in particular for adding/removing ingredients to/from a recipe.  For most of its subclasses, adding an Ingredient to a Recipe actually results in adding a copy of that ingredient, and creating a parent-child link between the new copy and the ‘master’ ingredient from which it was duplicated. This means we have to be a bit careful with (re)doing and undoing add or remove. I tried a couple of different approaches, but what seemed to be the most elegant was to make ‘add’ and ‘remove’ as symmetric as possible.  The Database class code for adding Ingredients already had the capability (via ‘noCopy’ parameter on various member functions) to add Ingredients with and without copying.  So I made a couple of changes:
- When you call ‘add’ for an Ingredient we would normally copy, it now works out whether or not it needs to make a copy of the ingredient by checking whether the ingredient has a ‘parent’ or not. To save doing copy-and-paste on the code in Database that looks for an Ingredient’s parent, I pulled it out into common functionality and added getParent() and setParent() member functions to Ingredient.  To save always going to the db, I also added member variable parentKey to store the key of the parent object.
 - When you call an ‘add’ or ‘remove’ function, you get back a pointer to a ‘what changed’ object, which is the object you’ll need to pass in to the inverse function.  So, if you call ‘add’ to a recipe and pass in the master copy of, say, a Fermentable, you’ll get back a pointer to the copy that was created of that Fermentable – because that’s what you’ll want to pass in to ‘remove’ to ‘undo’ the ‘add’.  Moreover, if you want to ‘undo’ and so call ‘remove’ with that child Fermentable, you’ll get back the same parameter you passed in.  If you then ‘redo’ the action, you’ll call ‘add’ but this time with the child Fermentable.  The ‘add’ function will work out that this child Fermentable already has a parent and will just re-store it in the DB without making a copy.  (The key of the child Fermentable will change, but that doesn’t break anything AFAICT.)
 - As part of implementing the above, I added the ‘insertInDatabase()’ virtual member function to Ingredient, which each subclass of Ingredient implements appropriately. In many places this does not make a big difference over the previous code of the caller (eg calling ‘m_target->insertInDatabase()’ instead of ‘Database::instance().insertWater(m_target)’), because the caller already knows what subclass of Ingredient it is dealing with.  However, in Database::addIngredientToRecipe(), it allows us to call ‘ing->insertInDatabase()’ without having to know what type of Ingredient we are adding.

As coded, if you remove several fermentables at a time, it creates and undoable action for each one.  If we don’t like this behaviour, it is possible to group undoable actions together.  We DO use that grouping functionality in MainWindow::droppedRecipeEquipment() for changes to equipment where the user has accepted to update batch size, boil size and time to that requested by the equipment.  Thus, if the user undoes the change to the equipment then the batch size etc changes are also reverted.  Using the Qt undo framework meant that this grouping was pretty much zero work to code. :)

I made some changes to MashStepTableModel to make it more like the other XxxTableModel classes – as that simplified keeping the GUI correct after undos/redos.  I suspect that, for some of the GUI updating, I have used callbacks where Qt signals could have done the same job.  Changing that might be part of a future refactoring.

I moved a comment from database.h to beerxml.h as it seemed to be related to the fromXml() function.

The changes to MainWindow::setupTriggers() are not as big as they look!  I added the new Edit menu (with Undo and Redo commands).  The other changes in here are just adding comments.

In some classes we have a convention that member variables are called m_foo, m_bar, etc, and in other classes not.  I’ve mostly tried to keep to an individual class’s style, but in new code I’ve added I mostly use ‘this->’ in front of access to a class’s own member variable.  I find it makes things clearer when reading the code, but tell me if you don’t like it.

Hope this all makes sense.  Give me a shout with any questions!

mikfire:
I am starting this review. Forgive me, but I need to clone this and run a few things over it. I promise I will either have comments or merge it by the end of the week.

Thanks for the work.

mikfire:
@matty0ung you say this is a partial implementation, which is cool. Do not follow my example and completely refactor the Database class in a single pull request.

As always, please understand I mean no personal attack or challenge here. I am simply poking at things and trying to clear up my confusion.

I am finding the behavior a little inconsistent. As an example, if I add some calcium chloride to a recipe and then change the amount from 0g to 5g, I cannot undo the change to the amount. The same steps on a fermentable allow me to undo the change to the amount but not the adding of the fermentable.

If this is expected and will be made more consistent in later PR, cool. If not, it needs to be addressed either here or in a later PR.

Otherwise, this seems at least as good as anything I would submit. I will give everybody until tomorrow to comment or object, and then I will merge.

matty0ung:
Thanks @mikfire.

I have now made changes to Hops/Yeast/Misc/Mash Step items in the recipe undoable, as these were relatively small changes.  (I suspect there are other places still to find where similar changes are needed, but I'm happy to fix those as and when we discover them.  I was reticent about, say, making changing ordering of mash steps undoable as it seemed like a bit of a chunkier change in what's already a sizeable pull request.  It should nonetheless be doable in future.)

It _should_ be the case that if you add a fermentable and then change the amount that you can undo changing the amount and then undo adding the fermentable.  This worked in my (admittedly limited) testing, but if there are circumstances where it doesn't then that's a bug.

mikfire (approved):

mikfire:
If I haven't said it, thanks for the work. This has been on the "want" list for a long time.

Issue #500, opened 2020-12-02
beerxml import seems broken


r4nd0m6uy:
Hallo Brewtarget developpers!

I have been trying importing my old beer recipes but it always crashes. Fixing one issue, another one shows up and I'm wondering if someone have tried this feature recently that seems totally broken. The following beer recipes always fail to import, despite they were exported with brewtarget:
https://gitlab.com/morandg/GuixBeers/-/tree/master/recipes

I don't know if I'm doing something wrong. I wish I could fix and PR myself but I'm getting lost in the issues from issues nightmare, any pointer is strongly appreciated.

Best regards,

Guy

r4nd0m6uy:
Last backtrace from original code I could get:
`
Thread 1 "brewtarget" received signal SIGSEGV, Segmentation fault.
0x00005555556b0032 in Ingredient::isValid (this=0x0) at /home/random/sources/brewtarget/src/ingredient.cpp:281
281	   return _valid;
(gdb) bt
#0  0x00005555556b0032 in Ingredient::isValid (this=0x0) at /home/random/sources/brewtarget/src/ingredient.cpp:281
#1  0x00005555556c219a in BeerXML::recipeFromXml (this=0x555555e9f600, node=...) at /home/random/sources/brewtarget/src/beerxml.cpp:1326
#2  0x0000555555755d2d in Database::importFromXML (this=0x555555e72e40, filename=...) at /home/random/sources/brewtarget/src/database.cpp:3051
#3  0x000055555580b8e4 in MainWindow::importFiles (this=0x555556188b40) at /home/random/sources/brewtarget/src/MainWindow.cpp:1944
#4  0x000055555572f18e in QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (MainWindow::*)()>::call(void (MainWindow::*)(), MainWindow*, void**) (f=(void (MainWindow::*)(MainWindow * const)) 0x55555580b7de <MainWindow::importFiles()>, o=0x555556188b40, arg=0x7fffffffd8d0)
    at /usr/include/qt5/QtCore/qobjectdefs_impl.h:152
#5  0x000055555572eb9c in QtPrivate::FunctionPointer<void (MainWindow::*)()>::call<QtPrivate::List<>, void>(void (MainWindow::*)(), MainWindow*, void**) (
[...]

PR #501, opened 2020-12-03
Fix for issue 500 - Copy being made of Database singleton

matty0ung:
This is a fix for a key issue in https://github.com/Brewtarget/brewtarget/issues/500 - though there may be other bugs to fix too.

BeerXML is making copies of the singleton Database object, which leads to strange bugs - eg it adds a style but the rest of the program can't see it because the in-memory reference to it in allStyles is only in the _copy_ of Database made by BeerXML.  BeerXML is able to make copies of Database, even though Database's copy constructor and copy assignment operator are private, because BeerXML is a friend of Database. 

The fix is to use the more modern approach to preventing copying - viz explicitly delete the copy constructor and copy assignment operator.  Compilation errors then tell you all the places in BeerXML where it is trying to copy the Database singleton (though actually they are easy to find as all the same format).  Change these copies to references and we're good.

### DIFF
--- a/src/beerxml.cpp
+++ b/src/beerxml.cpp
@@ -625,7 +625,7 @@ BrewNote* BeerXML::brewNoteFromXml( QDomNode const& node, Recipe* parent )
    QDomNode n;
    BrewNote* ret;
    QDateTime theDate;
-   Database db = Database::instance();
+   Database & db = Database::instance();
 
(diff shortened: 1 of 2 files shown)


r4nd0m6uy (approved):
Very nice!

mikfire (approved):

Issue #434, opened 2018-10-11
High DPI Display problems


jawsdaws:
![brewtargethdpi](https://user-images.githubusercontent.com/3501812/46827224-18b9fb00-cd66-11e8-968c-170ce391eb7f.png)
On my new HDPI (to me anyways) display there are some strange artifacts and general weirdness.  I've worked on some fixes.  They are at https://github.com/jawsdaws/brewtarget if anyone wants to take a look.  There's more to do though before I have a merge ready.

jawsdaws:
There looks to be an easy fix.  I'm going to do more testing over the weekend, and next week.  You can assign this one to me if you like.

jawsdaws:
I have some code for quite a few fixes here, but I'm not done.  I would like to get some help testing to make sure I'm on the right path.  I only have Linux available to test on so Mac and Windows help is needed.  Not sure the best way to get help on this though.  Maybe just merge the changes and see if it breaks?  I guess also if you see this and can help, checkout [https://github.com/jawsdaws/brewtarget/tree/HiDPI](https://github.com/jawsdaws/brewtarget/tree/HiDPI)

jawsdaws:
Thanks @DRiKE  . I tried that, but it was unreliable.  It appears to work for some display drivers, but not others.  I may be wrong, but that seemed to be the only thing I changed when it stopped working on Ubuntu 18.04.  I've some code to adjust some of the pixel values based on logicalDotsPerInch() and it seems to work and be more robust.  I've just been distracted with other things lately. :-)

dzsquared:
Just a note for anyone who comes across the high-DPI scaling issues on Windows - these are the properties changes to BrewTarget.exe that I've made to display the app
![image](https://user-images.githubusercontent.com/12227241/89969576-9c1b3080-dc0b-11ea-8582-69e48a09bed3.png)

Wyrmwood:
Thanks!

Regards,
Grognard

וצריך ביאור
*One way is to make it so simple that there are obviously no deficiencies,
and the other way is to make it so complicated that there are no obvious
deficiencies. - Tony Hoare*

dschwilk:
I have this issue with current develop branch build on kubuntu 20.04. Setting the environment variable QT_FONT_DPI=96 on the brewtarget command call does help.

jawsdaws:
[...]

PR #502, opened 2020-12-07
Highdpi

matty0ung:
Some fixes for https://github.com/Brewtarget/brewtarget/issues/434 - specifically for making the icons and range sliders look the right size, and for making the main text entry boxes large enough to show their contents.

Main advice for dealing with high DPI displays in Qt is either to query the screen resolution or to use the size of fonts that have been specified with screen-independent sizes (eg points rather than pixels). For things involving text, I have taken the latter approach, because, once you have made the widget the right size to show its text, you have won most of the battle.  For icons, I have used screen resolution to resize to an absolute size (eg 1/3 inch wide, etc).

 • In BtLineEdit.h and BtLineEdit.cpp, I have extended the BtLineEdit constructor to take an additional (but defaulted) parameter of an example of the longest text the creator wants it to display. We then use this to calculate, in calculateDisplaySize(), the width of the widget required to display this text. (AFAICT we don't need to mess with height because Qt already sets it to the right size for the font the widget is using.) It took a bit of investigation how to get all the elements to calculate the width, but this is explained in the comments. Finally, because setting in the constructor would be too early, we plug into the setText() methods to set the widget width we calculated in calculateDisplaySize().  And, by the way, if the widget's original width was greater than the minimum width we calculated, then we don't change it - ie we don't make the widget narrower, just wider if it would not otherwise show its text properly.
 • In MainWindow.cpp, I have added a new function, setSizesInPixelsBasedOnDpi(), that gets called from the constructor. This sets the sizes of the icons on the toolbar and the tab widget, and of the logo, by querying the screen for its dots-per-inch resolutions (horizontal and vertical) and then calculating pixel amounts from this for an absolute size.
 • Also in MainWindow.cpp, I have fixed the style-sheet used by the 'Boil SG' box so that it specifies the font size in points rather than pixels.  This means it should display the same size font on High DPI and regular monitors. 
 • Also in MainWindow.cpp, I fixed a bug with the IBU/GU slider where it displays insanely large values when OG is too close to 1.000.  Obviously this doesn't happen with a real recipe, but it is often visible when you have an incomplete recipe that you're working on.
 • In RangedSlider.cpp and RangedSlider.h, I have made the height of the slider dependent on the heights of the two bits of text it shows, and I have simplified the code that does all the drawing.  As mentioned in the comments, there probably is some slight further testing to do to see what happens when we drag the Brewtarget window from a High DPI screen to a regular one and vice-versa.  (I do not have such a set-up at the moment.)
 • In main.cpp, I have enabled AA_EnableHighDpiScaling, though it didn't seem to make any difference on my system (Linux laptop with High DPI screen) - so happy to drop this bit if it's of no benefit
 • In mainWindow.ui, I have fixed some missing tooltips and removed the unnecessary maximum widths on the range sliders (which otherwise is very notiecable on a high DPI monitor).

### DIFF
--- a/src/BtLineEdit.cpp
+++ b/src/BtLineEdit.cpp
@@ -26,8 +26,9 @@
 #include "Algorithms.h"
 #include <QSettings>
 #include <QDebug>
+#include <QStyle>
 
-BtLineEdit::BtLineEdit(QWidget *parent, Unit::UnitType type) :
+BtLineEdit::BtLineEdit(QWidget *parent, Unit::UnitType type, char const * const maximalDisplayString) :
    QLineEdit(parent),
    btParent(parent),
    _type(type),
@@ -36,6 +37,13 @@ BtLineEdit::BtLineEdit(QWidget *parent, Unit::UnitType type) :
 {
    _section = property("configSection").toString();
    connect(this,&QLineEdit::editingFinished,this,&BtLineEdit::onLineChanged);
+
+   // We can work out (and store) our display size here, but not yet set it.  The way the Designer UI Files work is to
+   // generate code that calls setters such as setMaximumWidth() etc, which would override anything we do here in the
+   // constructor.  So we set our size when setText() is called.
+   this->calculateDisplaySize(maximalDisplayString);
+
+   return;
 }
 
 void BtLineEdit::onLineChanged()
@@ -201,6 +209,8 @@ double BtLineEdit::toDouble(bool* ok)
 void BtLineEdit::setText( double amount, int precision)
 {
    QLineEdit::setText( displayAmount(amount,precision) );
+   this->setDisplaySize();
+   return;
 }
 
 void BtLineEdit::setText( Ingredient* element, int precision )
@@ -234,6 +244,8 @@ void BtLineEdit::setText( Ingredient* element, int precision )
    }
 
    QLineEdit::setText(display);
+   this->setDisplaySize();
+   return;
 }
 
 void BtLineEdit::setText( QString amount, int precision)
@@ -250,11 +262,16 @@ void BtLineEdit::setText( QString amount, int precision)
          Brewtarget::logW( QString("%1 could not convert %2 (%3:%4) to double").arg(Q_FUNC_INFO).arg(amount).arg(_section).arg(_editField) );
       QLineEdit::setText(displayAmount(amt, precision));
    }
+
+   this->setDisplaySize();
+   return;
 }
 
 void BtLineEdit::setText( QVariant amount, int precision)
 {
    setText(amount.toString(), precision);
+   this->setDisplaySize();
+   return;
 }
 
 int BtLineEdit::type() const { return (int)_type; }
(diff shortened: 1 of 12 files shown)


src/MainWindow.cpp
```diff
@@ -158,39 +160,104 @@ MainWindow::MainWindow(QWidget* parent)
...
+   //
+   // Setting height to be 1/3 inch seems plausible for the default size, but looks a bit wrong in practice.  Using 1/2
+   // height looks better.  Then width 265/66 × height.  (Note that we actually put the fraction in double literals to
+   // avoid premature rounding.)
+   //
+   // This is a bit more work to implement because its a PNG image in a QLabel object
+   //
+   Brewtarget::logD(QString("Logo default size: %1,%2").arg(this->label_Brewtarget->width()).arg(this->label_Brewtarget->height()));
+   this->label_Brewtarget->setScaledContents(true);
+   this->label_Brewtarget->setFixedSize((265.0/66.0) * dpiX/2, dpiY/2);
```
mattiasmaahl, line 254: @matty0ung 
I may be out in deep waters here swimming with the big boys :), but as I read this you only scale the X?
Should you not be scaling the Y as well?
i.e.  `this->label_Brewtarget->setFixedSize((265.0/66.0) * dpiX/2, (265.0/66.0) * dpiY/2);`

src/BtLineEdit.cpp
```diff
@@ -26,8 +26,9 @@
 #include "Algorithms.h"
 #include <QSettings>
 #include <QDebug>
+#include <QStyle>
 
-BtLineEdit::BtLineEdit(QWidget *parent, Unit::UnitType type) :
+BtLineEdit::BtLineEdit(QWidget *parent, Unit::UnitType type, char const * const maximalDisplayString) :
```
mikfire, line 31: If we need to internationalize this parameter (for example, something on the Style editor), will this char const*const work? Should we maybe consider a QString?

src/MainWindow.cpp
matty0ung: @mattiasmaahl Fair question.  In this instance, the code is (I believe!) correct but my comment could be better.  My intent was that the Y height should be half an inch - hence dpiY/2.  Then the ratio of width to height is 265:66 (from the dimensions of the original image).  So we multiply half an inch (dpiX) by 265/66.  I'll expand the comments a bit.

src/BtLineEdit.cpp
matty0ung: Ah, yes, good point.  Will fix.

matty0ung:
Thanks for comments.  Have made some changes in response.  Have also re-merged to latest develop, and fixed (c) comments to be in line with other files. (@mikfire I hope this last is OK, since you are original author of said files.)

mattiasmaahl:
Was reading through the code and compiling it on my laptop with a lower resolution, i.e. 1600x900, and found that if I ran the app in windowed form, not i maximized that is, the ranged sliders are a bil wonky. It's not that I think that it is a dealbreaker, but thought that I would bring it to your attension, earlier versions had a smaller slider which fit better at that windows size.
maybe make them a bit smaller?
As I wrote, NOT a dealbreaker, the app works just fine i maximized form, and I don't think I would use it in any other way any way.

Your thoughts?

brewtarget-resolution

mikfire:
It is one problem with the sliders. If I've understood this correctly, they are about the lowest level kind of canvas you can have in Qt5. This means the Qt engine assumes we will handle scaling, and it means that the rest of the engine really cannot make any guesses on how to scale around them.

I've seen this in other contexts. I am somewhat willing to let it roll as is, but maybe somebody with a hires monitor would like to see if we can fix the scaling on the sliders? Anybody else have input?

matty0ung:
Ah, interesting.  Good catch @mattiasmaahl.  By the looks of it, on your screenshot, there is not enough vertical space to display the sliders as they want to be.  But we could display them better in that space as we actually allow more vertical space than we strictly need for the green numbers.  The simplest thing would be to give only the space that the green numbers need, which may be a small change.  (I'll check.)  If that's not enough, then there are more complicated things we could do, where the slider reduces the sizes of the fonts it uses when it detects that vertical space is constrained.  That might not be a huge change either, but might not be necessary.  Will have a play and post a code update today or tomorrow.

mattiasmaahl:
@matty0ung maybe set the minimum height of the sliders so that when the windows gets that small you'll have to scroll in that section? If you make the windows smaller it already makes the section with the sliders scollable, I would suggest we make that section scroll earlier with bigger sliders? what do you think?

matty0ung:
@mattiasmaahl Yes, your solution sounds best.  I have uploaded some changes to do this.  (I forgot we were calling set minimum size too early, so it was getting overridden by the absolute pixel values from the UI file.  Have change the call so our calculated value takes precedence.)  I've also fixed a couple of other minor things (per commit comments).  Seems to work better now when I resize the window to something small, but would be great if you could have a quick look too.

mattiasmaahl (approved):
Tested on Debian 10 Buster and Windows 10 2004 looks nice!
both on 1600 resolution.

mattiasmaahl:
I find this PR to be in order, @mikfire get to have the final say in merging though.
I tested this on Debian 10 Buster on Laptop and on stationary Windows 10(2004) and both compiles and runs fine with latest fixes.

PR #503, opened 2020-12-08
Fix configure script

peterurbanec:
Prevent configure script from exiting early on non Debian/Ubuntu systems.

While at it clean up the mkdir invocation and redundant semicolons in the script.

### DIFF
--- a/configure
+++ b/configure
@@ -65,7 +65,7 @@ fi
 echo "CMAKEOPTIONS: $CMAKEOPTIONS"
 
 # Create dir only if needed
-mkdir build || true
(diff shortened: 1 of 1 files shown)


mattiasmaahl:
Thank you for your contribution.
I will go through the code tomorrow.

mattiasmaahl (approved):
Looks good

mattiasmaahl:
Looks good, although we usually don't do the in-source-build, rather we make a new build directory outside of the source directory and use cmake to configure the build there.

Issue #504, opened 2020-12-09
Make XML import more robust


matty0ung:
Following https://github.com/Brewtarget/brewtarget/issues/500, I'd like to make the XML import more robust:
- Current error-reporting when something goes wrong isn't as user-friendly as it might be - eg software exits with an error message to stdout...
- Some things that are, strictly, errors (eg missing mash step name) in the creation of the file are actually things that we could and ideally should recover from.  We perhaps can't auto-magically parse every erroneous BeerXML file, but we should IMHO target (a) handling common errors and (b) as far as possible importing any file that was exported from a current or previous released version of Brewtarget

mattiasmaahl:
I agree, we should be able to quite easily implement this, maybe just as a start report on the bigger categories, i.e. Mash-step, Fermentables, Equipment and such. or do you feel a more detailed error message is to be prefered?
My thinking here is to utilize the logging module more, maybe rewrite part of is so that it is always writing a Debug log file, but keeping it limited to the last 100 rows for example. this could also be used for issue reporting as well if need be?

matty0ung:
Yes, probably a pop-up window can give summary info about whether the import succeeded or failed - perhaps with first catastrophic error in the latter case - and then refer the user to the log file for blow-by-blow details.

mattiasmaahl:
I can tinker with the logging module and see if I can figure out a good strategy for the log file rotation and the dry-run-import. That is if you're not in the process already?

matty0ung:
Good idea.  I think having log file rotation is a good think independently of XML logging.  I tend to add lots of debugging logging when I'm working on a feature (and then delete some of it if it's insanely verbose), but it's relatively easy as a developer to create a large log file. :smiley: 

I am starting to have a look at the XML code.  Although @mikfire already did some good work to pull XML parsing out of the Database class, I think we can go further in decoupling.  I am also wondering if the import process cannot validate against BeerXML schemas (available at https://github.com/brewpoo/BeerXML-Standard/tree/master/schema) using QXmlSchemaValidator.  This would maybe save us reinventing some wheels to answer "is this valid BeerXML".

mattiasmaahl:
Sound god to me, will start working on the Log file rotation, will revert when I have something.

Regarding XML Import, I agree @mikfire did a good job on the XML parsing, and if the QXmlSchemaValidator works that would be a very good time-saver, but I have to read up on that library as I have never used it before.
Will revert when I have done some preliminary work.

[...]

PR #506, opened 2020-12-16
Rewriting the Log library to better suite the QT framework.

mattiasmaahl:
This PR will add new functionality to the Logging in Brewtarget.
Short description of what's been updated:
 - Log-Class has been rewritten as a Log Namespace to better suit the Qt Framworks QDebug library.
 - Refined logging and file management
 - Build the log rotation functionality
 - Build a test in QtTest for the logging librarys log rotation functionality.
 - Fixing up Copyright on modified files.

After these changes you can use either Brewtarget:logX(....) functions or use the QDebug- functions directly, i.e. QDebug() << "some nifty debug message" << someVariable;

This is part 1 of Issue #504 to help with logging when importing BeerXML files into Brewtarget

### DIFF
--- a/src/BtDigitWidget.cpp
+++ b/src/BtDigitWidget.cpp
@@ -64,11 +64,9 @@ void BtDigitWidget::display(QString str)
       display(m_lastNum,m_lastPrec);
    else
    {
-      Brewtarget::logW( 
-            QString("%1 : could not convert %2 to double")
+      qWarning() << QString("%1 : could not convert %2 to double")
             .arg(Q_FUNC_INFO)
-            .arg(str)
-      );
+            .arg(str);
       QLabel::setText("-");
    }
 }
@@ -175,7 +173,7 @@ void BtDigitWidget::setHighMsg( QString msg ) { m_high_msg = msg; update();}
 void BtDigitWidget::setMessages( QStringList msgs )
 {
    if ( msgs.size() != 3 ) {
-      Brewtarget::logW("Wrong number of messages");
+      qWarning() << "Wrong number of messages";
       return;
    }
    m_low_msg = msgs[0];
@@ -275,7 +273,7 @@ void BtDigitWidget::displayChanged(Unit::unitDisplay oldUnit, Unit::unitScale ol
       default:
          val = Brewtarget::toDouble(text(),&ok);
          if ( ! ok )
-            Brewtarget::logW( QString("%1: failed to convert %2 (%3:%4) to double").arg(Q_FUNC_INFO).arg(text()).arg(m_section).arg(m_editField) );
+            qWarning() << QString("%1: failed to convert %2 (%3:%4) to double").arg(Q_FUNC_INFO).arg(text()).arg(m_section).arg(m_editField);
          amt = displayAmount(val);
    }
    QLabel::setText(amt);
@@ -308,11 +306,11 @@ void BtDigitWidget::setText(QString amount, int precision)
    {
       amt = Brewtarget::toDouble(amount,&ok);
       if ( !ok ) {
-         Brewtarget::logW( QString("%1 could not convert %2 (%3:%4) to double")
+         qWarning() << QString("%1 could not convert %2 (%3:%4) to double")
                .arg(Q_FUNC_INFO)
                .arg(amount)
                .arg(m_section)
-               .arg(m_editField) );
+               .arg(m_editField);
       }
       m_lastNum = amt;
       m_lastPrec = precision;
--- a/src/BtLineEdit.cpp
+++ b/src/BtLineEdit.cpp
@@ -102,7 +102,7 @@ void BtLineEdit::lineChanged(Unit::unitDisplay oldUnit, Unit::unitScale oldScale
       default:
          val = Brewtarget::toDouble(text(),&ok);
          if ( ! ok )
-            Brewtarget::logW( QString("%1: failed to convert %2 (%3:%4) to double").arg(Q_FUNC_INFO).arg(text()).arg(_section).arg(_editField) );
+            qWarning() << QString("%1: failed to convert %2 (%3:%4) to double").arg(Q_FUNC_INFO).arg(text()).arg(_section).arg(_editField);
          amt = displayAmount(val);
    }
    QLineEdit::setText(amt);
@@ -160,7 +160,7 @@ double BtLineEdit::toSI(Unit::unitDisplay oldUnit,Unit::unitScale oldScale,bool 
    bool ok = false;
    double amt = toDouble(&ok);
    if ( ! ok )
-      Brewtarget::logW( QString("%1 : could not convert %2 (%3:%4) to double").arg(Q_FUNC_INFO).arg(text()).arg(_section).arg(_editField) );
+      qWarning() << QString("%1 : could not convert %2 (%3:%4) to double").arg(Q_FUNC_INFO).arg(text()).arg(_section).arg(_editField);
    return amt;
 }
 
@@ -231,11 +231,11 @@ void BtLineEdit::setText( Ingredient* element, int precision )
       // through toString() and then Brewtarget::toDouble().
       amount = tmp.toDouble(&ok);
       if ( !ok ) {
-         Brewtarget::logW( QString("%1 could not convert %2 (%3:%4) to double")
+         qWarning() << QString("%1 could not convert %2 (%3:%4) to double")
                               .arg(Q_FUNC_INFO)
                               .arg(tmp.toString())
                               .arg(_section)
-                              .arg(_editField) );
+                              .arg(_editField);
       }
 
       display = displayAmount(amount, precision);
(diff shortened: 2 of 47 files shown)


matty0ung:
This looks good - excellent work!  Really nice to have log file rotation and to be able to write `QDebug << "Foo";` etc.  From looking through the code, I had a couple of questions / minor suggestions, albeit that some are not specific to your changes:

 • For the mappings of LogType_DEBUG, LogType_INFO, etc to “DEBUG”, “INFO”, etc, could these be done via a static array?  Eg `char const * Log::logLevelNames {“INFO”, “DEBUG”, etc};`  Then `assert(QString(Log::logLevelNames[LogType_INFO]) == QString(“INFO”));` etc.  Similarly a couple of arrays could handle the mapping between `LogType_XXX` and `QtXXXMsg`.  It would save a bit of the boilerplate switches and if-thens, and would future enhancements easier (including extending log types and, maybe one day, even translating log messages).

 • In getLogFileList(), the comment says “checking that neither is bigger than the specified size restriction”, but I couldn’t immediately see this happening in the code.

 • Perhaps it's for a future enhancement, but could we one day make LOG_FILE_COUNT and LOG_FILE_SIZE configurable?  (Also, as an aside, do numerical constants such these need to be `#defines`?  Could they be `int const`?  See https://stackoverflow.com/questions/1637332/static-const-vs-define.)

 • Could we use `QMutexLocker` to do RAII and avoid having to do manual matching of `lock()/unlock()` calls on the mutex?

mattiasmaahl:
You make many good points! Will look into the above suggestions tomorrow and see what I can produce.
Thank you very much for hour input!

mattiasmaahl:
I absolutly agree that the LOG_FILE_COUNT and SIZE could be user setting, it's not that hard to implememt either, will fix tomorrow.

I have never used the QMutexLocker but will read up on it tomorrow and make the necessary ajustments. Should not be a big change as there is only two places where the log files are being written or deleted.

The comment above is left over from when I refactored the code at one point, it should be rewritten to better describe what is happening. 😊

matty0ung:
If you haven't used RAII for locks before, I think you'll like it.  It's super elegant IMHO, and I notice its absence when I have to do manual unlocking in languages that don't have RAII.

Instead of using `myMutex.lock();` and having to remember `myMutex.unlock();` on every exit path of the code (including exceptions), you just write `QMutexLocker lockContainer{&myMutex};` to acquire the lock and it is automatically released (by the QMutexLocker destructor) when lockContainer goes out of scope. So you can have a function with multiple returns, or that might throw exceptions, and "everything just works" (TM) with regard to mutex being released.  (And, of course, if you want the scope of the lock to be smaller than a function, just put `{` before acquiring the lock and `}` at the point you want to release it.  Your code is still automatically exception-safe with regard to the mutex.)

mattiasmaahl:
That sounds awsome, it's exactly like context managers in python and alike! I really dig those! Will definitely look into that tomorrow!

mattiasmaahl:
I've managed to implement the above suggestions except the static arrays for LogType, time wasn't allowed for that yet today. 
Will start working on it now, maybe I can manage to get something up today, otherwise I'll look into it tomorrow.
I'm sneaking in some development between work, some days there is much time to code other a little less.

@matty0ung If you could comment on the above commits to start with while I begin to code for the arrays.

mattiasmaahl:
@matty0ung Question, in the Brewtarget class there are these logging functions:
```
void Brewtarget::logW( QString message )
{
   if ( Log::loggingEnabled && Log::logLevel <= Log::LogType_WARNING)
      qWarning() << message;
}
```

As I'm rewriting the Log Library to handle all the checking an mangling of logging types and if there should be any logs written, do we really want these still??

I mean why would we type
`Brewtarget::logD("some message to log");  `
instead of
`qDebug() << "some message to log";`

my thinking here is that the brewtarget log functions do not bring any extra functionality to the program, just one more way to send logs, and being more clunky to use none the less!?

Any thoughts?

mattiasmaahl:
There, Have at it! go nuts! tell me what I screwed up! 😂

joking aside, please review as is now, made quite a few changes and looking forward to your comments and feedback!
Now the Enums are mapped using a char const* array to better translate between enum and string.
also made changes to check the user settings if the logs should be written and if so, to what level.

let me know what you think!

matty0ung:
Cool.  Will have a look.  Re `Brewtarget::logD` vs `qDebug()`, I agree - why re-invent the wheel, especially as (per https://doc.qt.io/qt-5/qtglobal.html#qDebug) qDebug gives you the choice between iostreams-style and printf-style output.  The question is whether you can bear to go through changing all the uses of Brewtarget::logD etc... :smiley:

matty0ung:
On the question about whether logging is enabled, I didn't check what we do at the moment, but I'd be very tempted _not_ to give users the option to turn logging off altogether.  There is an advantage to knowing that the app will always log important things (Errors and Warnings, but also build info) in a known/easily-findable place - eg when a user reports a bug, we often want to ask for the log file to get clues about what's going on.  Of course we could say "please try again with logging turned on" but this assumes the bug (a) is easily reproduceable by the user and (b) does not prevent the app starting up (so the user can easily enable logging through the UI).

mattiasmaahl:
Regarding the logD and friends, of course I can go through and change them all, feel that we shoild have one way of logging and then only have one place to maintain for it. Will have something up tomorrow for that.

Regarding enable/disable logging, when I wrote those settings I thought that the user could have a choice, and as I, at that point, did not have a way to control the file size I opted to have the user disable it by choice. Now that we can control diskusage I can refactor tje options to better suit the new logging library. Will look at it tomorrow!

mattiasmaahl:
I noticed that I had left behind some asserts from my debugging, fixing i a push today.

matty0ung:
Another thought - and apologies for the feedback coming piecemeal - is about filenames.  If I understood the code, we're logging to a file whose name is of the form "Brewtarget_Log_2020_12_18__09_07_23_123.txt" - set in `initLogFileName()` - with all the numbers being a timestamp set either when the app starts up or when we fill up a log file and need to move to the next one.  I see the benefit, that we have easily unique file names and you never need to mess around renaming files.  At the same time, I worry there might be a couple of downsides:
- Every time you stop and start the application, you get a new log file.  So, even if we set max log file size to something big, we might not actually retain that much logging if the max number of log files is small.
- It will be (slightly) harder to get users to submit log files because we have to explain to them what the names mean and how to find the right one.
I wonder if there might be some benefits to having the current log file always called "Brewtarget_Log.txt" and then saving the more complex names for the older files.  That way, log file rotation only happens when the log file is full, and users always know the name of the log file we want them to submit.  WDYT?

mattiasmaahl:
No worries, about the "piecemeal" comments, Commits are cheap! 

Interesting thought, and a was pondering the user of sequential numbering instead. 
for example, always writing to the Brewtarget_log.txt file, but on rotation remove the brewtarget_log_5.log, then rename the others as follows, figured it to be a bit clunky, but I never considered just naming the files with date when rotating and then always writing to the Brewtarget_log.txt, this is an easy fix, will have something up by the end of [my|the] day! 😊

mattiasmaahl:
This push is a Biggie.
replaced all logX( entries in all files with their corresponding QDebug version.

also fixed the logrotation to always write to Brewtarget_Log.txt and renaming them with timestamp when they reach the set size.
all new settings fot this is ready for implementation in the options dialog, but did not have the time today to fix that.

will continue with that in next week, maybe in a new PR? WDYT?

matty0ung:
Good work - thanks @mattiasmaahl.  Yes, I'll approve these changes to get them merged, then further tweaks can be in a separate PR.  (I'm reasonably comfortable that the basic logging functionality will get a fair bit of testing during development.  We'll just want to be sure that we haven't missed any edge cases on rotation etc.)

matty0ung (approved):

jimdbr:
Is anyone else seeing a build error with this version?

brewtarget/src/Log.cpp:190:47: error: explicit qualification in declaration of ‘QString Log::getTypeName(Log::LogType)’
  190 |    QString Log::getTypeName(const LogType type) {

mattiasmaahl:
> 
> 
> Is anyone else seeing a build error with this version?
> 
> brewtarget/src/Log.cpp:190:47: error: explicit qualification in declaration of ‘QString Log::getTypeName(Log::LogType)’
> 190 | QString Log::getTypeName(const LogType type) {

Good catch jimbr! fixing in a new PR, there is a typo in the declaration that I missed. PR Coming up!

mattiasmaahl:
PR #510 adresses the above bug.

matty0ung:
Some other problems to fix.  I'm getting the following when I run Brewtarget with this new code:
`QDir::mkpath: Empty or null file name
"Log is in a temporary directory: /tmp/Brewtarget_Log.txt"
[08:56:23.600] INFO : "XDG_CONFIG_HOME not set.  HOME directory is /home/matt", in 404`
Can we default to the same log file location as previous versions of the code?

mattiasmaahl:
hmm, that is wierd, did it not find the correct path to make the logging directory in?

I'll have a look at it. To the best of my memory, I did not make that big of a change to the Linux side of directory making, more on the Windows side of things, due to the fact that the program found the wrong directory for APPDATA, but I'll have a look at it and see if I can fix it today.

mattiasmaahl:
@matty0ung I have a new PR for the above issue with the Log file not being able to initialize that early in the process.
Please see #511

matty0ung:
By the way, https://github.com/Brewtarget/brewtarget/pull/506/ is a good example of why we shouldn't ignore compiler warnings.  The warning about brewtarget.cpp:1217 is actually a bug (missing `qWarning() <<` before QString).  Can you include fix in your patch?

mattiasmaahl:
I saw the above, but forgot to include it, will fix tomorrow.

PR #510, opened 2020-12-21
Declaration of getTypeName was incorrectly declared in Log.cpp. fixed.

mattiasmaahl:
As jimdbr so correctly found is that the getTypeName was incorrectly defined in Log.cpp.
This PR fixes that issue.

Built and testes on both windows and Linux.

### DIFF
--- a/src/Log.cpp
+++ b/src/Log.cpp
@@ -187,7 +187,7 @@ namespace Log
       //emit wroteEntry(logEntry);
    }
 
-   QString Log::getTypeName(const LogType type) {
+   QString getTypeName(const LogType type) {
       return QString(logLevelNames[type]);
    }
 
--- a/src/Log.h
+++ b/src/Log.h
@@ -118,6 +118,6 @@ namespace Log
 
    /* Get the list of Logfiles present in the directory currently logging in and returns a FileInfoList containing the files.*/
    extern QFileInfoList getLogFileList();
-};
+}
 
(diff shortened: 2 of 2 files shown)


matty0ung (changes requested):
Good stuff.  Could you also fix compiler warnings - eg extra ; on line 121 of Log.h.

mattiasmaahl:
Done! 😊

matty0ung (approved):

jimdbr:
Builds and runs fine now. Still a couple of compiler warnings.

mattiasmaahl:
@jimdbr sounds good, yes there are a few compiler warnings, but there is a fair amount of development going on right now on the development-branch, clean-ups will happen at some point later. If you're interested, we're always lokking for help?

matty0ung:
Are the compiler warnings easy to fix?  Always nice to stay on top of these things if we can. :smile_cat:

mattiasmaahl:
It's a few unused db pointers here and there, can have a lokk to see if I can incorperate them as well in the current pr #511

jimdbr:
I'm an old School Fortran and C coder so my skills are pretty rusty, I am happy to build, test and report bugs.

matty0ung:
Oh, yes, you're right: a couple of unused variables in beerxml.cpp.  I'm not seeing it in my copy because I already fixed it on my local branch. :smile:   OK, that's not a big deal.

PR #511, opened 2020-12-21
Fixing Log initialization.

mattiasmaahl:
Now it starts to log to file after options has been read.
Early logging is sent to stderr, and Logging is enabled by default.

Error was thrown when Linux did not supply a valid path for log file at initialization, this was changed to only log to stderr at startup.
On windows there was a path supplied on my machine at least, but not on my linux laptop.

### DIFF
--- a/src/BtTreeItem.cpp
+++ b/src/BtTreeItem.cpp
@@ -314,7 +314,7 @@ QVariant BtTreeItem::dataMisc(int column)
             return QVariant(misc->useStringTr());
          break;
       default :
-         QString("BtTreeItem::dataMisc Bad column: %1").arg(column);
+         qWarning() << QString("BtTreeItem::dataMisc Bad column: %1").arg(column);
    }
    return QVariant();
 }
--- a/src/Log.cpp
+++ b/src/Log.cpp
@@ -53,11 +53,17 @@ namespace Log
    };
 
    // options set by the end user.
+   // Default logging set to enable to get the first logs out.
    bool loggingEnabled(true);
-   LogType logLevel = LogType_DEBUG;
+   // Default logging level set to INFO, to get only status logs.
+   LogType logLevel = LogType_INFO;
+   // Stores the path to the logfiles
    QDir logFilePath;
+   // Set to true if you want brewtarget to figure out where to put the logs.
    bool logUseConfigDir = true;
+   // Set the log file size for the rotation.
    int const logFileSize = 500 * 1024;
+   // set the number of files to keep when rotating.
    int const logFileCount = 5;
    // \brief this is the file we're always logging to.
    QString logFileName("brewtarget_log");
--- a/src/Log.cpp
+++ b/src/Log.cpp
@@ -60,7 +60,7 @@ namespace Log
    int const logFileSize = 500 * 1024;
    int const logFileCount = 5;
    // \brief this is the file we're always logging to.
-   QString logFileName("Brewtarget_Log");
+   QString logFileName("brewtarget_log");
    // \brief what file ending to use for the file.
    QString logFileNameSuffix("txt");
    QString timeFormat;
--- a/src/Log.cpp
+++ b/src/Log.cpp
@@ -53,8 +53,8 @@ namespace Log
    };
 
    // options set by the end user.
-   bool loggingEnabled = false;
-   LogType logLevel = LogType_INFO;
+   bool loggingEnabled(true);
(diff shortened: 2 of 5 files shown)


matty0ung:
Can you talk me through how we get to logging to the right place after we've started logging to stderr?  I started reading through the code, but got confused by there being two changeDirectory() functions - one that takes a "defaultDir" parameter (but doesn't seem to get called?) and one that doesn't take any parameters.

matty0ung:
Could we make another couple of tweaks to make this a less surprising upgrade for users?  Default log file name has changed from "brewtarget_log.txt" to "Brewtarget_Log.txt".  Can we change it back?  Also, it was not respecting a previously set log level and switched the setting to INFO.  Did we change the name of the parameter or something?  If so, again, can we change it back?

mattiasmaahl:
Let me look into those issues tomorrow, the name is easy to change, that is just a parameter.
Getting from stderr to file is done by setting the Log::logFilePath and then call Log::changedirectory()

The Log::changedriectory( defaultdir ) function is not in use anymore, that should be removed, I haven't gotten around to it yet.

Let me get back to you on the others tomorrow.😊

mattiasmaahl:
After looking at this some more, I cannot see that the settings was not respected. 

> 
> 
> Could we make another couple of tweaks to make this a less surprising upgrade for users? Default log file name has changed from "brewtarget_log.txt" to "Brewtarget_Log.txt". Can we change it back? Also, it was not respecting a previously set log level and switched the setting to INFO. Did we change the name of the parameter or something? If so, again, can we change it back?

@matty0ung, question on the second part, How do you mean when you say It does not respect the settings? When I was looking in to this I found that the combobox was not displaying the correct level when you go in to change the settings although the setting was saved in the config file.

I fixed this in my code to be pushed, I also changed the name on the log file to lower_case name. added a bunch of comments to try to explain how the logging is initialized, but in short, the logging is initialized to start logging to the stderr/console on startup, then when the app options is read the settings for the logging is updated to start logging to file. this results in that the first logs is always in the console and never lands in the log file, this is not optimal, and I'm trying to figure out a way to get all the logs saved into the final log file somehow, maybe start logging in an temp-file and then copy/move the file to the new location once the location is known? WDYT?

mattiasmaahl:
> 
> 
> Can you talk me through how we get to logging to the right place after we've started logging to stderr? I started reading through the code, but got confused by there being two changeDirectory() functions - one that takes a "defaultDir" parameter (but doesn't seem to get called?) and one that doesn't take any parameters.

in short, the logs are initialized in the Brewtarget::initialize() function, there it starts out logging only to stderr/console. later in the same function when the users/app settings are applied the Log::changedirectory() is called to set the log file and start a textstream to save logs to a file and Console if this is the users preference.
In Brewtarget:initialize() the function readSystemOptions() is called, this will add the settings to the Log namespace but they will not be applied to the logging unless you call the Log::changeDirectory() function. I made it this way because I did not want to have a state in the logging library where somthing was half way set and logging was written.
I'm sure there are better ways to do this, but will require further development down the road. For now I'm trying to fix my f**k-ups! 😂

mattiasmaahl:
Sorry, @mikfire I miss clicked, did not mean to "request a review" from you! 😂😊

matty0ung (approved):
Looks good - many thanks!

matty0ung:
Sorry for slow replies on my part.  Been a bit busy with a few things going on here that have a hard deadline of 24 Dec! :smiley: 
Thanks for explanation and additional comments.  I agree it makes sense to enable logging to stdout/stderr for the time before we have read in config.
Re the log levels, what happened was that I had set my log level to DEBUG a long time ago and then it went back to being INFO after I merged v1 of the new logging code.  I thought perhaps we'd changed the name of the config parameter or something, but maybe it was just that there was a version of the code that wasn't finding the config file.  Anyway, after I changed it to DEBUG again, it hasn't reverted, so it's OK now.

mattiasmaahl:
No worries on the slow replies! this IS an open source, pro bono if you will, project! there are no requirements to reply in a specific time! 😊

I do value yours and @mikfire replies and comments as they develop and hone my skill in the language, I mostly dabbled in C# and Python, so the whole C++ experience is an old skill not practiced for a long time for me! And I have always been (very) bad at commenting my work, so there is always room for improvement! 
And to be honest, this is the first project I've been involved in that uses the QT framework! 😊 (loving it so far!)

I wish you all a Merry Christmas and a Happy New Year! 🎅

PR #515, opened 2021-01-25
Fix for making XML import more robust

matty0ung:
This is a patch for https://github.com/Brewtarget/brewtarget/issues/504

Before testing this code, please make a manual backup of your database.  This is for two reasons.  Firstly, there is still a chance that if you hit a bug part-way through an import, your database can end up in an unreadable state. (Ask me how I know.)  Secondly, if you want to test importing the same file multiple times, it’s useful to be able to revert the database to a prior state.

You’re also going to want to install Xerces and Xalan.  (See comment in xml/XmlCoding.cpp about why we need to use Xerces and Xalan over the Qt built-in XML features.  Essentially the Qt features are being partly withdrawn because they were not compliant and users are now encourage to look at third party libraries.)  This is pretty simple to do on Ubuntu:
```
$ sudo apt update
$ sudo apt install libxerces-c-dev
$ sudo apt install libxerces-c-doc
$ sudo apt install libxalan-c-dev
$ sudo apt install libxalan-c-doc

```

**What’s changed**
There are two main parts to this patch.  One is to add XSD validation as a preliminary stage to reading in BeerXML files.  The other is to change the way we read in XML, in no small part so that we can report errors (and success) to the user.  The new way of reading in XML is also more data-driven and generic, which should make it easier to read in other XML files in future if we want to.  In particular, I have tried to decouple XML processing from the database implementation.  The new code instantiates objects (eg Hops, Yeasts, Recipes, etc) from the XML and then asks those objects to save themselves in the DB.

I have not, for now, touched saving XML, though it would be an obvious next step to extend the new XML reading code to do this, which would then sever the tight coupling between the BeerXML and Database classes (that is left over from the fact that BeerXML was separated out of Database).  I thought saving was lower priority as it’s inherently less error-prone because we do not (or at least should not) have to validate all the data we are processing.  We can trust the data in our DB, whereas we cannot know what errors might be contained in some random file we’re trying to read in.

Suggest you first look at `brewtarget/xsd/beerxml/v1/BeerXml.xsd`.  This is an attempt to create an XSD for BeerXML v1 documents.  As mentioned in a comment near the end, there’s a small hack we have to do in code to turn BeerXML into legal XML, but apart from that, using this XSD does filter out a lot of errors that we would otherwise hit in processing a BeerXML file.  (Amusingly it finds real errors in some of the sample files downloadable from beerxml.com.)

On the C++ source code, I was worried that we were getting to have a lot of files in the src directory, so I thought to start creating some structure.  Most of the new code is in the src/xml directory.  (You will see I have also created src/model and moved and renamed `ingredient.h` to `NamedEntity.h` in that directory.  In time I would propose to move other ‘model’ classes, eg hop.h, hop.cpp, recipe.h, recipe.cpp, etc to this directory too.)

For the C++ code, there are three things going on:

- Loading in the XSD schema – see `XmlCoding::impl::loadSchema()`.  This was a lot of the pain in getting started with Xerces, for reasons explained in the comments.  Once you’ve got this bit working, the rest of the APIs for Xerces and Xalan – or at least the bits of it we need to use – are pretty straightforward.

- Validating an XML document against the XSD schema – see `XmlCoding::impl::validateLoadAndStoreInDb()` along with helper classes `BtDomDocumentOwner`
 (just an RAII wrapper), `BtDomErrorHandler` (for logging errors and deciding whether they are ignorable), `XercesHelpers` (not strictly needed but useful for logging all Xerces settings), `XQString` (which handles such things as conversions between Qt’s Qstrings and Xerces/Xalan strings)

- Loading in an XML document – see `XmlCoding`, `XmlRecord`, `XmlNamedEntityRecord`, `XmlRecipeRecord`, `XmlMashStepRecord` and `XmlRecordCount` (which keeps track of statistics).  This new code tries to keep generic XML processing separate from the specifics of an XML coding.  The “mapping” for BeerXML is now mostly just data in the first 500 lines or so of `beerxml.cpp`, which should make it easier to maintain.  (There is also a small amount of code in `BeerXML::impl::validateAndLoad()` to do the XML hack for BeerXML and set up “acceptable XML errors in BeerXML documents” before passing off to generic code in `XmlCoding::validateLoadAndStoreInDb()`.)

A few other things along the way include moving `Database::importFromXML()` to `BeerXML::importFromXML()` and some minor tidy-ups.

I've spent a bit more time on this than I expected I would need, largely because it took a few attempts to work out a way of nicely separating the generic and encoding-specific parts of the code.  This meant I did a few refactorings and it's possible (despite my efforts to the contrary) that not all of the comments have kept up with this.  So please do shout if you see any comment that is unclear or in any way does not match the code!


xzfantom:
I think it would be good idea to mention new libs in readme file, too

matty0ung:
Have updated the README per @xzfantom's suggestion.

xzfantom:
Finally I managed how to add new libraries (though, I still have troubles building xalan, but found binaries. I'll update instructions later).
When building project I got this error:
```
[build] C:/Qt/5.15.2/mingw81_64/include/QtCore/qlist.h: In instantiation of 'void QList<T>::node_construct(QList<T>::Node*, const T&) [with T = std::pair<const char* const, std::shared_ptr<XmlRecord> >]':
[build] C:/Qt/5.15.2/mingw81_64/include/QtCore/qlist.h:625:13:   required from 'void QList<T>::append(const T&) [with T = std::pair<const char* const, std::shared_ptr<XmlRecord> >]'
[build] C:/Qt/5.15.2/mingw81_64/include/QtCore/qhash.h:1254:13:   required from 'QList<T> QMultiHash<K, V>::values(const Key&) const [with Key = const QString&; T = std::pair<const char* const, std::shared_ptr<XmlRecord> >]'
[build] E:\Projects\brewtarget\src\xml\XmlRecipeRecord.cpp:42:44:   required from 'void XmlRecipeRecord::addChildren() [with CNE = Hop]'
[build] E:\Projects\brewtarget\src\xml\XmlRecipeRecord.cpp:72:27:   required from here
[build] C:/Qt/5.15.2/mingw81_64/include/QtCore/qlist.h:471:35: error: use of deleted function 'std::pair<_T1, _T2>& std::pair<_T1, _T2>::operator=(typename std::conditional<std::__not_<std::__and_<std::is_copy_assignable<_Tp>, std::is_copy_assignable<_T2> > >::value, const std::pair<_T1, _T2>&, const std::__nonesuch_no_braces&>::type) [with _T1 = const char* const; _T2 = std::shared_ptr<XmlRecord>; typename std::conditional<std::__not_<std::__and_<std::is_copy_assignable<_Tp>, std::is_copy_assignable<_T2> > >::value, const std::pair<_T1, _T2>&, const std::__nonesuch_no_braces&>::type = const std::pair<const char* const, std::shared_ptr<XmlRecord> >&]'
[build]      else *reinterpret_cast<T*>(n) = t;
```

matty0ung:
Ah, interesting.  Windows compiler must be a bit different than gcc.
I've changed the bit of the code that gives that error so it's not trying to copy the values.  It (still) compiles and runs on Linux.  If you get a minute, could you see if it now compiles on Windows?

That sounds cool about Xalan too.  (Personally, I'm not too worried about building things like Xalan and Xerces from source as long as the source and reliable binaries are readily available.)

xzfantom:
Yes, it's working now for mingw in windows.
I wouldn't bother building this libraries, too, but in windows there is no apt :) On official web-site no binaries, either. I found project [MSYS2](https://www.msys2.org/) that makes building apps in  windows same as in linux world. Only thing is that folder with all toolkit (Devtools + QT5 + CMake + libs) is about 3.5Gb - not really a problem, but looks like over engineering for me.

xzfantom:
One more note: in BeerXML standard field DATE is described "Date brewed in a easily recognizable format such as “3 Dec   04”. Is it possible to accept this format, too?
[brauhaus.zip](https://github.com/Brewtarget/brewtarget/files/5903096/brauhaus.zip)

matty0ung:
Glad it's working now.

I hadn't spotted that they only have 32-bit binaries for Windows Xalan.  That's a bit annoying.  (It's been a good few years since I used Xalan and Xerces on Windows, and even then it was only as a dev environment for something that ran on Solaris in production, so back then I think I just used the 32-bit version for dev and the 64-bit version in prod, none of which helps us now.)

Will have a look at the date thing.

matty0ung:
OK, I've extended the date code to accept more different formats.  Your file now loads.  (Qt thinks the two-digit date in it is 1904 rather than 2004, but fixing that would require adding our own Y2K heuristics, which, unless others have strong views, I'm a little reluctant to do as it seems such heuristics are bound to always be wrong in some circumstances.)

Greatly appreciate your testing and feedback.  Keep it coming! :)

xzfantom:
I discovered that Xalan-C++ 1.12.0 is [not compatible with Cygwin and MinGW](https://github.com/apache/xalan-c/releases/tag/Xalan-C_1_12_0), but MSYS2 added some [patches](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-xalan-c). So, I'll play around it a little bit and decide way for autobuild

xzfantom (approved):
Tested on Windows 10, looks good

xzfantom:
Finally I managed how to autobuild your new code in github actions for windows with help of msys2. I didn't change anything in code tree, so it builds only binaries, not installer yet. Not a big deal, but I have some thoughts about installer that can make building it easier.

xzfantom:
I can't start app after last commit. It builds, but not starting. And I can't start debug, too. Does it start for you?

matty0ung:
Oh, how odd.  Yes, it all ran just fine for me.  Are you getting any output at all in `brewtarget_log.txt`?

xzfantom:
> 
> 
> Oh, how odd. Yes, it all ran just fine for me. Are you getting any output at all in `brewtarget_log.txt`?

Nope, it's always empty. I wanted to play with this problem later. On my ubuntu box I start test with no screen:
```
env CTEST_OUTPUT_ON_FAILURE=1 env QT_QPA_PLATFORM=offscreen make test
...
Config: Using QtTest library 5.12.8, Qt 5.12.8 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 9.3.0)
brewtarget_tests: xercesc/util/XMemory.cpp:63: static void* xercesc_3_2::XMemory::operator new(size_t, xercesc_3_2::MemoryManager*): Assertion `manager != 0' failed.
...
```
I'll test today at work, may be can get smth else.

xzfantom:
So it looks like xerces & xalan are initialized in main.cpp, but QTests uses it's own main for test. So the way is to move xerces init to Brewtarget::initialize() or copy&paste it to testing.cpp

matty0ung:
Looks like it's an initialisation problem in the test suite.  In the app, in `main()`, we initialise Xerces:
```
   // Initialize Xerces XML tools
   // NB: This is also where where we would initialise xalanc::XalanTransformer if we were using it
   try {
      xercesc::XMLPlatformUtils::Initialize();
   } catch (xercesc::XMLException const & xercesInitException) {
      qCritical() << Q_FUNC_INFO << "Xerces XML Parser Initialisation Failed: " << xercesInitException.getMessage();
      return 1;
   }
```
But test has its own `main()`, so doesn't run this.

matty0ung:
Simplest fix is probably to copy and paste this code into Testing::initTestCase().  I'll have a look.

matty0ung:
Tested locally and the following additions to Testing.cpp are a fix:

`#include <xercesc/util/PlatformUtils.hpp>
`

At start of Testing::initTestCase():
```
   // Initialize Xerces XML tools
   // NB: This is also where where we would initialise xalanc::XalanTransformer if we were using it
   try {
      xercesc::XMLPlatformUtils::Initialize();
   } catch (xercesc::XMLException const & xercesInitException) {
      qCritical() << Q_FUNC_INFO << "Xerces XML Parser Initialisation Failed: " << xercesInitException.getMessage();
      return;
   }
```

At end of Testing::cleanupTestCase():
```
   //
   // Clean exit of Xerces XML tools
   // If we, in future, want to use XalanTransformer, this needs to be extended to:
   //    XalanTransformer::terminate();
   //    XMLPlatformUtils::Terminate();
   //    XalanTransformer::ICUCleanUp();
   //
   xercesc::XMLPlatformUtils::Terminate();
```

Will push this change once https://github.com/Brewtarget/brewtarget/pull/524 is reviewed (as I already merged locally).

xzfantom:
Looks good now

xzfantom (approved):
Builds on win, lin & mac
[Test action](https://github.com/xzfantom/brewtarget/actions/runs/568523975)
Manual test on Win 10

mikfire (commented):
If you can rebase this, I will merge it. Sorry for the extra work -- I did things in the wrong order

matty0ung:
Thanks @mikfire. Have now merged in latest changes from HEAD.

mikfire (approved):

Issue #518, opened 2021-01-27
Creating a new recipe makes the program crash on start-up


matty0ung:
Steps to reproduce (obviously make sure you have a backup of your database first!):

- Start Brewtarget
- Right-click on the Recipes tree pane and select New > Recipe
- Type in a recipe name, eg 'Bork', and click OK
- Close Brewtarget
- Start Brewtarget again ... result = core dump

If you want your data back, edit the DB in another tool to remove the Recipe record you just added.  Brewtarget will now start OK again.

I'm presuming the issue is that when creating a new Recipe object, one (or more) of the fields is set to a default value that is not valid.  Looking at the DB, it seems that a lot of fields are set to NULL.

Hopefully the fix is something along the lines of creating some sensible default values for fields in Recipe constructor.

matty0ung:
Hmm, actually, on investigating further, this seems to be a problem of initialisation order.  Somewhere in the sequence of things that happens inside the MainWindow constructor, we're ending up at `RecipeExtrasWidget::updateDate()`, which is calling `Brewtarget::mainWindow()` to get a pointer to the MainWindow object but getting back null because MainWindow constructor hasn't returned to `Brewtarget::run()` yet so `Brewtarget::_mainWindow` is not yet set.

This is partly a consequence of the Undo/Redo code needing to have all "updates" done via MainWindow.

I think the fix will be to have less going on in the MainWindow constructor - just initialisation of its essential member variables - and move some of the other set-up to an init() function that is called after the constructor returns.

PR #519, opened 2021-01-29
Stop crash on start when new Recipe added

matty0ung:
Fix for https://github.com/Brewtarget/brewtarget/issues/518

It turns out code called from the MainWindow constructor was relying on the return value from the MainWindow constructor...


mattiasmaahl (approved):
Looks good!

Issue #520, opened 2021-01-30
Put string constants for property names in namespaces and match names to values 


matty0ung:
I have been thinking more about the global constants ktableFoo, kpropBar, and so on.  At the moment they are a bit double-edged.  The benefit is that, if you refer to, say, `kpropAmountKg` in the code, you can be sure you've not amount_kg the "amount_kg" property name it refers to (which would be an embarrassing run-time error).  The downside is that, if you're looking for all the places we refer to the "amount_kg" property, then you have an extra layer of indirection.  To take a different example, if you want to find all the places we refer to equipment ID, you might end up doing the following:
```
$ grep -r equipment_id ../src
../src/database.cpp:      // Update equipment_id
../src/TableSchemaConst.h:static const QString kpropEquipmentId("equipment_id");
../src/TableSchemaConst.h:static const QString kcolEquipmentId("equipment_id");
../src/RecipeSchema.h:static const QString kcolRecipeEquipmentId("equipment_id");
$ grep -r kpropEquipmentId ../src
../src/database.cpp:   int id = get( tbl, parent->key(), tbl->foreignKeyToColumn(kpropEquipmentId)).toInt();
../src/database.cpp:                QString("%1=%2").arg(tbl->foreignKeyToColumn(kpropEquipmentId)).arg(newEquip->key()),
../src/TableSchemaConst.h:static const QString kpropEquipmentId("equipment_id");
../src/TableSchema.cpp:   m_foreignKeys[kpropEquipmentId] = new PropertySchema( kpropEquipmentId, kcolRecipeEquipmentId, QString("integer"), Brewtarget::EQUIPTABLE);
$ grep -r kcolEquipmentId ../src
../src/TableSchemaConst.h:static const QString kcolEquipmentId("equipment_id");
../src/TableSchema.cpp:         defineBtTable(kcolEquipmentId, Brewtarget::EQUIPTABLE);
$ grep -r kcolRecipeEquipmentId ../src
../src/RecipeSchema.h:static const QString kcolRecipeEquipmentId("equipment_id");
../src/TableSchema.cpp:   m_foreignKeys[kpropEquipmentId] = new PropertySchema( kpropEquipmentId, kcolRecipeEquipmentId, QString("integer"), Brewtarget::EQUIPTABLE);
$ 
```
[...]

PR #524, opened 2021-02-14
Rename property name constants to use namespaces and match names to values 

matty0ung:
This is a fix for https://github.com/Brewtarget/brewtarget/issues/520
I wrote a script to do the changes, which I can share if folks are interested, though obviously it's the changes themselves that count.


mattiasmaahl (approved):
Builds, and looks good.

mattiasmaahl:
As it is a "small" change, I'll merge it.

Issue #403, opened 2018-03-05
Brew it / generate brew notes


pricelessbrewing:
Soooo I didn't even know this existed and I've been using brewtarget for awhile. @mikfire  

I'm going to suggest making one of these tabs visible at all times, and that the "date" text field change the name of the tab? Or maybe change the tab name from a date to a sequential number system, so brew 1, brew 2 etc? That would make it more visible IMO.

pricelessbrewing:
Alternatively, have that as a button along the toolbar? 

Honestly I couldn't figure out where this was again today and had to go look on the issue Mikfire wrote out on beerjson repo...

matty0ung:
I do agree this is a bit of a hidden feature.  I only discovered its existence from working on the code.  Will have a think about where to put an additional more visible way to access it than the current right-click on a recipe.

matty0ung:
I've added a fairly visible button for this functionality - see https://github.com/Brewtarget/brewtarget/pull/525.
Happy to hear views about whether the button should be in a different place or have a little icon instead of text etc.

PR #525, opened 2021-02-14
Add a Brew It button to recipe

matty0ung:
This is a fix for https://github.com/Brewtarget/brewtarget/issues/403 to make the "brew notes" functionality more visible (ie not just accessible from a right-click on the tree of recipes).


mattiasmaahl (approved):
Small enough change, builds fine. I'll approve it.

Issue #526, opened 2021-02-14
Brewtarget doesn't stop you running multiple instances


matty0ung:
Feels like we should check at start-up whether another instance of the program is already running.  At the moment, on Linux, if you start a second instance then the first instance is locked out of the DB (assuming you're using the default settings and thus SQLite) which means it crashes (sometimes with an error message and sometimes without) when you try to, say, edit or create a recipe.

Unless we explicitly want to allow people to run two or more instances of the program simultaneously, I suggest we keep things simple and disallow it.  Plenty of other programs do this, and it saves us having to worry about cross-process race conditions, deadlocks etc.

PR #527, opened 2021-02-16
Prevent multiple simultaneous instances of Brewtarget

matty0ung:
Fix for https://github.com/Brewtarget/brewtarget/issues/526.  Based on one of the simpler 'standard' approaches published on StackOverflow.


mattiasmaahl (approved):
Good upgrade I think. I'll approve.

PR #530, opened 2021-02-27
Add radar chart to Edit Water dialog

matty0ung:
I was starting to look at water profile tools and I thought a nice first step might be a visualisation, so I created a simple radar chart.


mikfire (approved):
I don't suppose there is any way we could re-use the schema constants for the property names? One of the intents of that refactoring was to make it easier to add a field by only having to edit one location instead of having to grep through the code and find every place that needs modification.

This otherwise looks good, and I will approve the merge. We can discuss the constants later.

PR #536, opened 2021-03-31
Closes #532 -- Problem with amounts in inventory

### LINKED ISSUE #532: Problem with amounts in inventory
I found interesting feature in inventory management. I added new malt - it was the same type from different supplier. So I copied the existing to the new one. Entered new amount and it changed in old and new one (source and copy). It is the same with hops. Is it a bug or new feature?

mikfire:
The code was working as designed, I just had a bad design. I had missed
this use-case, and so the original's inventory_id was copied along
with the rest.

The fix was to add a boolean to the new* calls in Database that allowed
us to specify if we wanted a new inventory row or not. Since this
allowed me to declutter some of the code, I think it is a reasonable
approach.

The only path I could find that needed this was from
BtTreeModel::copySelected, so that method got modified to request the
new inventory row. I had considered putting the code to add a new
inventory row here, which had the benefit of keeping the complexity
close to where it was used. But I didn't like the model having to open a
transaction, so I put into the Database methods.

My quick testing makes me think this is working better.


mattiasmaahl:
I was reading through the code and tried to build it on my Debian, and it failed due to a function/method not being defined.

Question: Where can I find the function/method newIngredient(...)?

matty0ung:
This is probably a merge issue.  It's now called newNamedEntity().  (Not all Ingredients were ingredients, if you see what I mean, so I changed the name of the class to NamedEntity as part of the XML work.  There's a comment in the code in model/NamedEntity.h that elaborates a bit more.)

mattiasmaahl:
Aha, ok. Now that you're saying it I do remember that discussion/PR. then it's an easy fix for Mik to do before merge. :smile:

src/database.cpp
```diff
@@ -1479,35 +1479,33 @@ Equipment* Database::newEquipment(Equipment* other)
...
+      if (other != nullptr) {
          tmp = copy(other, &allFermentables);
       }
       else {
-         // new ingredients don't. this gets ugly fast, because we are now
-         // writing to two tables and need some transactional protection
-         sqlDatabase().transaction();
-         transact = true;
-         tmp = newNamedEntity(&allFermentables);
+         tmp = newIngredient(&allFermentables);
```
mattiasmaahl, line 1493: Should be: 
tmp = newNamedEntity(&allFermentables);

mattiasmaahl (changes requested):
All the newInventory(...) need to be changed to newNamedEntity(...) before merge

mikfire:
Ugh. Missed that part. I'll update after work.

mattiasmaahl (approved):
Builds and runs.

PR #549, opened 2021-04-08
Mac package

xzfantom:
#545 Building mac package in native environment


mattiasmaahl (approved):
I do think we need to have this verified to work on a Mac, But will approve for now and let us test on a Mac when possible.

Issue #456, opened 2020-04-06
CI/CD automated builds


cgspeck:
It would be good to have CI/CD and automated binary builds on PRs and merges.

I suggest the goals of this would be to:
1) automatically build binaries for popular linux targets ~~supported by Docker~~
2) automatically run the tests
3) automatically publish ~~(as [Github releases](https://github.com/cgspeck/brewtarget/releases))~~ passing binaries
4) publish a Guacgui docker images [with Brewtarget develop installed](https://hub.docker.com/r/cgspeck/brewtarget-guacgui)

It could in the future be extended to create:
a) OS specific dev releases (e.g. update PPA for Ubuntu, AUR for Arch)
b) Windows builds (it would be amazing to be able to recommend this to my non-tech friends).

I've created a similar workflow using TravisCI on my fork and offer it as [an example](https://travis-ci.com/github/cgspeck/brewtarget/builds/136039274). 

I will have some spare time coming up soon and will be happy to work on this in a few weeks.

xzfantom:
Why this issue was closed? I think the idea of CI/CD is very good and was working in same direction. And as far as I can see @cgspeck already made most of it.

matty0ung:
Looks like a typo - AFAICT intent was to close https://github.com/Brewtarget/brewtarget/issues/455 rather than https://github.com/Brewtarget/brewtarget/issues/456.  Reopening.

cgspeck:
I have opened a draft PR with my working build, see #523 

What specific version of QT we should be using (I chose 5.14.2, it needs to be >= 5.10 for the test suite)?

Should we explicitly list Debian dependencies or rely on `CPACK_DEBIAN_PACKAGE_SHLIBDEPS`? (see [this patch](https://github.com/Brewtarget/brewtarget/pull/523/files#diff-1af3ff0c718f9f5b8b422a01a148a5761adb2f7dcb66a7e0d38114a1b98a9bf1))

xzfantom:
About minimal QT5 version we have issue #521 and there is no final decision yet.

I was working on the same question here [develop.yml](https://github.com/xzfantom/brewtarget/blob/feature-cpack-lintian-check/.github/workflows/develop.yml). And some my thoughts are here:

I used MSYS2 because it has it's package manager and library xalan-c, that can be useful if PR #515 get merged.

[...]

Issue #521, opened 2021-02-02
QT5 minimal version


xzfantom:
In CMakeLists.txt QT5_MIN_VERSION is declared as 5.3. It's outdated value and current project can't build with QT older than 5.10.
Looks like we should update this value to more reasonable one.
And one more thing - in Ubuntu 18.04 LTS in packages QT5 is [5.9.5](https://packages.ubuntu.com/bionic/qt5-default). Do we need to support it?
@mikfire

matty0ung:
From the dates at https://en.wikipedia.org/wiki/Qt_version_history it seems like 5.12 would be a sensible minimum.
OTOH, if a lot of end users are likely to already have old versions of Qt installed, maybe we should just static link against a fixed version of Qt?

xzfantom:
I think that as far as we don't use features of new QT releases we should use as low version as we can. For example in current Debian Buster QT is 5.11, so using 5.10 will allow this people to use BT with no additional effort. As for me, I'd prefer static linking only if using new QT version will bring us real benefits. 
I'm not good in QT development, so I don't have reasonable opinion here.
Maybe somebody else can say something on this topic.

matty0ung:
Yes, you're right we should make things as easy as possible for users.  And dynamic linking is generally preferable for lots of reasons.  I guess I was just thinking about static linking as a fall-back if we can't rely on a new enough version of Qt being installed.  Eg, AIUI, we need at least 5.11 for High-DPI Support, so I'd worry a bit about anything older than that.

As an aside, I can see a few places in the code that seem to do _compile-time_ checks against Qt version (`grep -ir "QT_VERSION" ../src` returns 11 results), but I haven't dug into any of these so I don't know if any of the should cause concern.

xzfantom:
> As an aside, I can see a few places in the code that seem to do _compile-time_ checks against Qt version (`grep -ir "QT_VERSION" ../src` returns 11 results), but I haven't dug into any of these so I don't know if any of the should cause concern.

It's to hide some stuff deprecated in new QT versions if new feature is not introduced in old version. Should be no problem, most checks are with 5.15, some with 5.13.

matty0ung:
Ah, OK, cool.

xzfantom:
Looks like we have some consensus about QT 5.9.5 in LTS Ubuntu 18.04. Some discussion in this pr #550.

PR #550, opened 2021-04-08
CI win + xalan

xzfantom:
As mentioned in #456 one of the working ways to build app for windows is using MSYS2. To avoid finding required libs for different build systems for Windows I added some code to CMakeLists.txt to find it automatically.

### DIFF
--- a/.github/workflows/linux-ubuntu.yml
+++ b/.github/workflows/linux-ubuntu.yml
@@ -13,8 +13,10 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-18.04
-
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, ubuntu-latest]
     steps:
       - uses: actions/checkout@v2
         with:
@@ -23,7 +25,15 @@ jobs:
       - name: Dependencies
         shell: bash
         run: |
-          sudo apt-get install -y \
+          sudo apt-get update && sudo apt-get install -y \
+            qtbase5-dev \
+            qttools5-dev \
+            qttools5-dev-tools \
+            qtmultimedia5-dev \
+            libqt5sql5-sqlite \
+            libqt5sql5-psql \
+            libqt5svg5-dev \
+            libqt5multimedia5-plugins \
             lintian \
             rpmlint \
             libxerces-c-dev \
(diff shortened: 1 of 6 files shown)


mikfire (commented):
Could I talk you into reducing the required Qt5 version to 5.9? 

The Qt5 packages for Ubuntu 18.04LTS are all Qt 5.9.5. We need to maintain support for that version until it drops into extended maintenance mode.

mattiasmaahl:
I may be wrong here, but doesn't QT5.9 break the Test-package for the build?

xzfantom:
> 
> 
> I may be wrong here, but doesn't QT5.9 break the Test-package for the build?

I checked this like few months ago and minimal version for project to build was 5.10. We had open issue on this question here: #521 
I'll check one more time just to be sure.

mikfire:
No need. I've just checked. It does break on 5.9. 

I also verified that you can't build on 18.04LTS with the minimum version of 5.12, at least using the default bionic packages.

xzfantom:
`/home/runner/work/brewtarget/brewtarget/build/src/brewtarget_tests_autogen/EWIEGA46WW/../../../../src/Testing.h:70:22: note: suggested alternative: ‘randomStringGenerator’
          int index = QRandomGenerator().generate64() % posChars.length();
                      ^~~~~~~~~~~~~~~~
                      randomStringGenerator`

QRandomGenerator() was introduced in QT 5.10 only

xzfantom:
> 
> 
> No need. I've just checked. It does break on 5.9.
> 
> I also verified that you can't build on 18.04LTS with the minimum version of 5.12, at least using the default bionic packages.

Maybe add condition and change QRandomGenerator on qrand()?

mikfire:
Ok. I can fix this. A few `#if QT_VERSION` tricks like we are already using through the code base.

If you will set the minimum version to 5.9 and break the tests for a bit, I will submit a PR to fix this issue. I may not be generating cryptographically secure random numbers, but I don't think we need them in this case.

mikfire:
@xzfantom I yield to your solution, and will keep my hands off.

xzfantom:
Added matrix build for Linux, so it's QT5.9.5 and QT12 check

mikfire (approved):

PR #555, opened 2021-04-11
Bugs/constructors

mikfire:
Side discussions made me aware that I had horribly broken the model in the primitive constructors by making each of them aware of the database column names. This is my first pass at fixing it.

I was able to remove a lot of unused constructors, one slightly redundant template method and generally clean the code up. It comes with a price tag of a slightly more complex NamedEntity constructor, and that one of the newNamedEntity methods has to search for the record it just created.


matty0ung (approved):
Looks good. Definitely an improvement that the constructors can now treat QSqlRecord as a "black box" from which properties can be extracted by name.

PR #561, opened 2021-04-14
Minor fixes / tidy-up to XML mappings

matty0ung:
I found a typo in the XML mappings (my bad copy-and-paste), so have fixed and also made sure we're using PropertyNames constants throughout.  Changed CMake min version to 3.16 so I could compile on my system (see discussion at https://github.com/Brewtarget/brewtarget/discussions/559).

### DIFF
--- a/src/beerxml.cpp
+++ b/src/beerxml.cpp
@@ -100,27 +100,27 @@ namespace {
       {"Leaf",   Hop::Leaf}
    };
    XmlRecord::FieldDefinitions const BEER_XML_HOP_RECORD_FIELDS {
-      // Type              XPath             Q_PROPERTY           Enum Mapper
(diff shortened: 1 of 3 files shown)


mikfire (approved):

src/beerxml.cpp
```diff
@@ -317,79 +317,79 @@ namespace {
       {"Infusion",     MashStep::Infusion},
       {"Temperature",  MashStep::Temperature},
       {"Decoction",    MashStep::Decoction}
-      // Inside Brewtarget we also have MashStep::flySparge and MashStep::batchSparge which are not mentioned in the
+      // Inside Brewken we also have MashStep::flySparge and MashStep::batchSparge which are not mentioned in the
```
xzfantom, line 320: Forgot to rename back

matty0ung: D'oh!  Sorry, will fix.

PR #575, opened 2021-04-22
Bugs/btstringedit

mikfire:
This touches a few files, but it fairly simple to explain.

The core of it was that the high DPI work was not recalculating the size of the edit field. This works in most places, as we have 3 - 5 digits, a decimal place and then 3 digits. I simply caused the size to be recalculated on a BtStringEdit when new data was added. To keep this within reason, I decided that we would draw nothing less than 8 characters (unicode) and nothing more than 50.

When I went to fix the various editors, things broke. The hop editor, in particular, kept getting resized and looked awful. 

All the editors are now tabbed forms. I've tried to move what I consider important to the main page, while leaving the rest to the back pages. Depending on what you like, this will either be an improvement or absolutely horrible.

Test this one carefully. It won't impact your database, but make sure the tabbed editors work. If they don't, we can discuss what would work.

### DIFF
--- a/src/BtHorizontalTabs.h
+++ b/src/BtHorizontalTabs.h
@@ -1,0 +1,57 @@
+/*
+ * BtHorizontalTabs.h is part of Brewtarget, and is Copyright the following
+ * authors 2021-2025
+ * - Mik Firestone <mikfire@gmail.com>
+ *
+ * Brewtarget is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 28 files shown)


mattiasmaahl (approved):
I like the new look, makes it easy to add functionality later if we need to, adding a new tab for whatever and off we go.
I cannot find anything that bugs out when I run the code, So I will approve this but hold off with merging until Matt has had a chance to have his say/opinion. 😊

PR #583, opened 2021-04-29
Tidy up of ABV tool, plus move remaining model class files into subdirectory

matty0ung:
This is a follow-on to https://github.com/Brewtarget/brewtarget/pull/367

Although a lot of files are touched, mostly this is just to move the other Ingredient/NamedEntity object files into the model subdirectory and make their naming consistent with other classes (Hop.h rather than hop.h etc).  The main files to look at are AlcoholTool.* and Algorithms.*

Since Algorithms was a class with only static member functions, I converted it to a namespace.  I fixed the layout on AlcoholTool so it works on high DPI displays.

There is still some minor bug with calls to BtDensityEdit::toSI() generating `WARNING : QSettings::value: Empty key passed, in 0` but I _think_ this is unrelated to the changes here.


mattiasmaahl (approved):
Looks good to me. will approve an merge.

PR #584, opened 2021-05-06
Undo works everywhere

mikfire:
Adds the necessary to make dropping ingredients from the trees onto a recipe undoable.

There are a few use cases I am simply ignoring. Since a multi-add can only happen when an ingredient is dropped on a recipe, I don't worry about finding parents. By definition, we have to make a copy when this happens.

I probably made a mess of the undo class. I considered making a different undo class for lists of things, but was concerned we would end up with two undo queues. It isn't pretty, but it does work.


mikfire:
Well, that went sideways. Let me try this again.

PR #586, opened 2021-05-09
Ideas/versioning

mikfire:
# Motivation
I write all my own recipes, and spend a lot of effort modifying them, experimenting with different yeasts, hop schedules, etc.

Currently, the only way I can preserve the historical data is to make a copy of the recipe and then modify the copy. My current database is littered with "Beer v1", "Beer v2", "Beer v4", etc. I skipped "beer v3" because I forgot to make the copy and now cannot recover what changes I made for v3.

This is exactly what software is supposed to do -- automate repetitive tasks so I don't have to waste brain space remembering to do them.

*WARNING*: Running this code will modify your database schema and bump the version from 9 to 10. Test this only with proper backups.

# Git, for brewing
This PR implements a basic versioning system for recipes. The user can choose an automatic system, where a modification to an already brewed recipe (aka, there is a brewnote associated with the recipe) immediately spawns a version.

The default is the user can right click on any recipe, select ancestors->version recipe to create the same effect.

The two systems are not mutually exclusive.

## Added bonus
As part of doing this, I had to be able to mark a recipe as read only. It took a simple effort to allow the user to mark any recipe as read only. 

# How it works.
The heart of it is a new column on the recipe table called `ancestor_id` and a recursive SQL query.

When recipe B is spawned from recipe A, `B.ancestor_id = A.id`. If A has no ancestors, then `A.ancestor_id = A.id`. Spawning a descendant will also mark the ancestor as read-only, and set display to false.

To build the list of ancestors, a recursive SQL query is used to find each ancestor id and stops when `ancestor_id == id`.

There is a lot subtly in there, including discussions of what deleting a recipe should do, orphaning recipes, etc. You can read more in dev-doc/recipeversioning.markdown.

## Copies. Lots and lots of copies
Of course, but "spawn" I mean "copy". And by "copy", I mean _everything_ -- hops, fermentables, miscs, etc. They all get copied when a new version is created.

The copy mechanism is actually somewhat interesting. See commit 6b9e8d32 and 22cf7ee2 for more details.

# What I need
Testing. I've tried the edge cases I can think of but I am not really creative like that. Tell me what works, what doesn't work, what needs fixed.


mattiasmaahl:
First off!! I really like this feature!

I have been playing around a bit, and had a thought. What if you create a recipe, then version 2 and version 3, but then want to create a new branch from version 2 or even the original?
I can make a copy of either of them and keep their respective tree/history, but have to restart BT for them to show ( bug! ), but I cannot create a new "version" of an ancestor in the tree. Is that by design? 
Should I use the "Copy"-function instead of the "Version"-function to keep the workflow separated? From what I can see the "branching" is working when you copy a recipe and the history is intact. 

More of a workflow question with a small bug report. 😊

mikfire:
One of my basic design choices was no branching.

As soon as we get branches, we will have to implement a diff/merge. I am really not that smart, or maybe I really do not want to reimplement git :laughing: 

If we were to change our minds and allow branching, we need some way to display the information and the relationship between all the branches. We would also need some way to show which descendants were active. It was more than I was willing to attack, and so I took a shortcut and said "no branching". 

My intent was that to go back to an ancestor, you would have to orphan each of its descendants. You could then use the ancestry tool to put the orphans back into their own tree.

The copy thing is a bug. My expectation is that you shouldn't be able to do that whatsoever, and that copying a recipe wipes out its ancestry. As you suggest, this might provide an interesting idea for how to implement a branch. Let me think on this.

mattiasmaahl:
I really like the idea of this, as I almost always work in a test-as-you-go way. and History is always nice to have if you "break" a recipe and tweak it to much.

Will test further and revert if I find something interesting.

matty0ung:
I do like the idea of versioning recipes.  It's a feature I would use.
For the terminology etc, I wonder if it's worth getting the input of someone who is not a software developer?  Personally, I know exactly what you mean when you talk about ancestors and descendants and branching and orphans etc, but I worry that the average brewer will  find such explanations too complicated.  My instinct would (however things work under the hood) be to present things to the user in somewhat the way that wiki page versioning is done (eg see https://weblearn.ox.ac.uk/portal/help/TOCDisplay/content.hlp?docId=howdoiviewwikipagehistory).

mikfire:
@mattiasmaahl  you actually found two bugs that had been lurking in the code for a while. Try the latest rev, although you may want to start with a clean database.

Ever think our data model and database interactions are far more complex than required?

mattiasmaahl:
@mikfire cool that the bugs where found and fixed! I will restore my database to default and try again.

As for translating "Orphan Recipe", Maybe you want to "Detach Recipe" or "Make independent"?

matty0ung:
I thought some more about the naming etc.  I like the idea of sticking to "version" as it's intuitive.  I think we could solve some of the other problems by extending the list of restrictions a little. How about the following for a list of what users can/cannot do:
- As now, "No branching".  As you say, it's too complicated.
- Only latest version is normally visible/brewable
- Old versions are not editable or deleteable
- If you want to "revert" to an older version you actually take a copy of it which becomes the latest version.  Eg, if Chocolate Lager v4 turned out to be a bust and I want to go back to v3 for my next round of tweaking, then we'll make Chocolate Lager v5 as a copy of v3.  (This is how wiki page versioning works.)
- As now, copying any version of any recipe just makes a new recipe with the same data as that version.  None of the ancestry is copied.  And the new recipe does not remember/know where it came from.
- If you want to "branch" off an older version, you can make a completely new (ie unrelated) recipe that's a copy of it

These rules mean, amongst other things, you would never have orphans.

mattiasmaahl:
I've been testing a bit and I'm happy with the functionality, do you feel you want to do more work on this, @mikfire, before merging?

mikfire:
Needs some more work. I found a bug that I *think* is a signaling issue, but hang me if I can find it.

If you want to look at it, do this:
   1. Enable automatic versioning on the options dialog (this matters);
   2. Brew a recipe
   3. Modify the recipe
   4. Observe that the leaf recipe doesn't show your changes, but the immediate ancestor does
   5. Restart brewtarget
   6. Observe the leaf recipe now shows the proper changes.

Given that the data does load correctly, this has to be something at the display layer. I just haven't been able to track it down. Until I do, this maybe shouldn't be merged?

There's also a bunch of things I want to rework a little -- name changes, clean up the code, etc.

mikfire:
Assuming the checks clear, I think this is in shape to merge. It has all the basic functionality I want and I can't find any obvious bugs. Anything else, I will fix as a problem ticket.

mattiasmaahl (approved):
Looks very good, I will merge

Issue #560, opened 2021-04-14
Boil time for hops is shown in whole hours (90 min -> 2h)


Alge:
When adding hops, the time is shown rounded to 2 hours instead of 1.5.

When I enter the time:
![Screenshot from 2021-04-14 13-51-48](https://user-images.githubusercontent.com/7383768/114706101-d869d800-9d28-11eb-90a9-030119accbe6.png)

After I press enter:
![Screenshot from 2021-04-14 13-51-53](https://user-images.githubusercontent.com/7383768/114706100-d7d14180-9d28-11eb-997c-b1e74b71b806.png)

On the preview:
![Screenshot from 2021-04-14 13-52-41](https://user-images.githubusercontent.com/7383768/114706099-d738ab00-9d28-11eb-9047-9fdca43b1cb5.png)

xzfantom:
Hello! What version do you use? Try download latest build from Actions: https://github.com/Brewtarget/brewtarget/actions and check.
![изображение](https://user-images.githubusercontent.com/5746986/114750570-cc980900-9d5c-11eb-965f-10d42e5d6975.png)
Be aware it's betta, but close to release.

Alge:
I'm using 2.3.1 downloaded yesterday. I'll try out the beta!

Alge:
Confirmed that it works in the beta! Always nice when issues are solved before reporting ;)

![Screenshot from 2021-04-15 09-18-33](https://user-images.githubusercontent.com/7383768/114829649-a3629180-9dcb-11eb-8e98-44ff3abfe284.png)

Alge:
Found another place where the time is rounded up!
Using 2.4.0, when showing the brewday print preview the time is still shown as 2 hours
![Screenshot from 2021-05-06 16-58-17](https://user-images.githubusercontent.com/7383768/117320388-78c89d80-ae8c-11eb-92a2-eaba42a7bc95.png)

matty0ung:
I'm getting minutes rather than hours for my fields in print preview, but it might be that the recipe I'm testing with doesn't have the problem fields.  Could you export your recipe to XML and then add that XML file to this ticket?

Alge:
Sure!

Couldn't add the file in the comment, so here's a pastebin link: https://pastebin.com/XMV5xvwD

Alge:
and here is a complete picture of the print preview:
[...]

PR #587, opened 2021-05-09
Tidy up Unit, UnitSystem classes etc

matty0ung:
It seemed to me that these classes were more complicated than they needed to be, especially with modern C++ features such as lambdas available to us.  This tidies them up and, in the process, fixes https://github.com/Brewtarget/brewtarget/issues/560.

Look at Unit.h, Unit.cpp, UnitSystem.h, UnitSystem.cpp first and then all the other consequential changes hopefully make sense.

Essentially, I've got rid of all the subclasses of Unit and UnitSystem because they were repeating code unnecessarily.  The differences between different units and unit systems are just data and (simple) formulae, so they don't need a plethora of subclasses.  I also removed/fixed some comments that seemed incorrect (eg about the iteration order of QMap).

I think there is still room for some further tidying up of UnitSystem (see comments in that class), but hopefully this is already something of an improvement.


mattiasmaahl:
Beautiful coding, really admire your coding style!

mattiasmaahl (approved):
Looks good and is beautifully refactored into a more idiomatic code, easy to read and understand.

PR #590, opened 2021-05-11
Quickfix with null-check on mash when copy recipe

### LINKED ISSUE #589: Brewtarget  (2.4.0) crashes when copying recipe
Steps to reproduce:

1. create a new empty recipe
2. right-click -> copy
3. brewtarget crashes

I used the .deb from the latest build (from today, 2021-05-11)

mattiasmaahl:
Close #589 
Null checking added to addToRecipe (mash) as new recipes do not have mashes by default the copy mechanism crashed when trying to handle a nullptr.

Quickfix waiting for a possible rework further down the road.

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -3110,6 +3110,9 @@ QList<Hop*> Database::addToRecipe( Recipe* rec, QList<Hop*>hops, bool transact )
 
 Mash * Database::addToRecipe( Recipe* rec, Mash* m, bool noCopy, bool transact )
 {
+   if ( m == nullptr )
+      return nullptr;
+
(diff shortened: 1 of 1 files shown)


matty0ung (approved):

matty0ung:
I'm assuming this doesn't break anything anywhere else when it returns nullptr... :)

mattiasmaahl:
I was thinking that maybe doing it as if the mash m = nullptr then just create a new mash, but there is no new mash created when you create a new recipe so I went with just returning a nullptr instead. I could not find any other crashes due to this but that's no guarantee. ☺

Issue #588, opened 2021-05-11
XML recipe import not setting amounts


matty0ung:
I've spotted a bug in the XML import of recipes.  The amounts of Hops, Fermentables, etc are not always getting saved.  I know why this is happening and I have a fix in mind.

PR #591, opened 2021-05-12
Fix missing amounts on some XML recipe imports

matty0ung:
This fixes https://github.com/Brewtarget/brewtarget/issues/588


mattiasmaahl (approved):
Looks good, as always! will approve

PR #592, opened 2021-05-20
new icon for brewit

mikfire:
I had never actually intended to leave the tool bar with the words "Brew It". I had wanted to put an icon in its place, but what sort of icon would be meaningful? And, just as important, free enough for us to use?

So I made this. I won't say "drew", because I really mostly copy/pasted other resources we already had. I did, however, add my own numeric 2.


mattiasmaahl (approved):
Nice icon!

PR #597, opened 2021-05-26
Partial solution to #596

mikfire:
The problem as seen in #596 was directly related to how I reordered the cache and flush.

Changing an existing hop (or anything else) on a non-versioning recipe would cause the NamedEntity:changed() signal to fire before the cache value was updated. Since everybody uses the cache value, nothing would change until we had reason to reload the recipe and re-read the values.

I am somewhat trapped for a solution on this. The order is very specific and versioning requires brewtarget clone before it updates the cache. It simply doesn't work any other way.

This means brewtarget must wait until after the clone decision is made to emit a NamedEntity::changed signal. The cache has to be updated before we signal and that can only be done in the derived class (eg Hop) and it cannot be done in NamedEntity.

This has had a kind of largish knock on effect, as every (almost) set* method in (almost) every major class had to change. BrewNote, Salt and Water escaped the carnage entirely -- none of those can fork a recipe.  Inventory changes on all the classes which support inventory also escaped, as changing inventory does not fork

I had to clean up Database a little, as the alternate findParentRecipe methods weren't actually being called. I removed those and made the idea work.

The last pain was that modifying a recipe's equipment profile was not quite working. I am not certain that changing something on the equipment profile should fork a recipe, but it does for now. The hard part was that the first field updated would fork the recipe, which would signal MainWindow which would reload the Equipment editor which would reset all the fields. My solution was to get the field values before we start writing. It mostly works. It is ugly.

The equipment thing may work better if I enable cacheOnly, and then use generateUpdateRow() to blast the entire thing record at once. That will neatly solve the ugly and the signal problem. It will require I make a clone-aware method for doing it, but I've done worse.

### DIFF
--- a/src/EquipmentEditor.cpp
+++ b/src/EquipmentEditor.cpp
@@ -221,27 +221,48 @@ void EquipmentEditor::save()
          return;
    }
 
-   obsEquip->setName( lineEdit_name->text(), obsEquip->cacheOnly() );
+   // this is one idea for handling how clones work. It is a very brute force
+   // idea, and I am really not sold on it yet?
+   QString name           = lineEdit_name->text();
+   double boil_size       = lineEdit_boilSize->toSI();
+   double batch_size      = lineEdit_batchSize->toSI();
+   double tun_volume      = lineEdit_tunVolume->toSI();
+   double tun_weight      = lineEdit_tunWeight->toSI() ;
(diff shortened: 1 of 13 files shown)


mattiasmaahl (approved):
Looking good. Aproving.
Any "ugly" parts can be fixed at a later point if need be! 😁

PR #600, opened 2021-06-24
Feature/printpreview

mattiasmaahl:
Print Preview.... THIS IS A DRAFT! 
I need input/sugestions on styling :blush:
I have with this made it WYSIWYG to print recipes and BrewDay Instructions. ( Inventory is not implemented yet, but it will come next week I think.)

So I wrote a markdown of the BtPage object that I wrote to handle the plasing and ordering of the text/objects on the page.
I have also modified the RecipeFormatter somewhat to get all the information in a format that is easy to handle and generate page content from. I will add the markdown text below, and it can also be found int the PRs files under the dev-doc folder.

now This code is not adding any Style so to speak to the printout, and what I mean by that is there is currently no way to add for example every other row grey background and so on.
This I would like to get some input.

And as always!! please comment and tell me if there is better ways to solve the problems in the code so I can learn!

# PrintPreviewDialog
This dialog is an attempt to get the printing to be *cross-platform-WYSIWYG* in that it uses the QT standard way of generating the printout and preview.
I also aimed to make a unified print/export dialog instead of having multiple menu items saying print-this and print-that.

## nBtPage namespace
I created this namespace to keep averything in one bubble of code and separate from the rest of Brewtarget. this was to minimize the risk of running into duplicate nameing.
This namespace includes all that is necessary for generating the printout and Html-export to file, while relying on Brewtargets RecipeFormatter- and BrewNoteFormatter-class to get the data for generating the pages. I'm not sure this is the way it should be in the future but will leave it there for now.

### BtPage class
This class is used to store all child elements/classes and placing them on the page. this class also contains the referens to the printer in use and is responible for calling all childrens ::render() method to get everthing printed onto the page.
This is genereated everytime you change your selection for printout or the prewview needs to refresh for any reason.
#### Funtions:
- **template \<class T> auto addChildObject(T \*obj, QPoint position = QPoint()) -> decltype(obj)**\
You add content to the page by creating and adding PageChildObject-subclasses, see below. this returns a pointer to the added object if you need to make more setting to is, i.e. adding a style to a table
- **renderPage()**\
Renders the page on the set printer/painter
- **template \<class T, class S> void placeRelationalTo(T \*targetObj, S \*sourceObj, PlacingFlags place, int xOffset = 0 /\* pixels \*/, int yOffset = 0 /\* pixels \*/)**\
Sets the postition of targetObj relational to the sourceObj according to the placingflags. you can also offset the postion using the Offset arguments.
- **template \<class T, class S> void placeRelationalToMM(T \*targetObj, S \*sourceObj, PlacingFlags place, int xOffset = 0 /\* Millimeter \*/ , int yOffset = 0 /\* Millimeter \*/)**\
Same as placeRelationalTo(..) but offset is in millimeters and this will calculate the distance in pixels according to the set printer and then call the original function with those values.
- **template \<class T> void placeOnPage(T \*targetObj, PlacingFlags place, int xOffset = 0, int yOffset = 0)**\
This will set the position on the supplied targetObj according to the PlacingFlags provided. this funtion is a convinient way of placing anyting to, for example, the TOP-RIGHT corner of the page or why not the center of the page. more on [PlacingFlags](#placingflags) below.
### PageChildObject
This is a abstract / Interface class for all child object to put on the page. It holds some basic informations about the object and makes sure that all derived classes implement a few methods, i.e. a render method, because all children need to know how to draw it self onto a QPainter and ultimatly onto a page.\
This class also contains information on the boundingbox and position that is the same for all derived classes.

#### Derived Classes
These are the derived classes for now, in the future I'll implement more children to draw more kind of objects on the page, for example arcs or circles or lines. I'm contemplating to have the MainWindows rangeWidgets print out on page, I think that would look good, Not htere yet though. :)

##### PageText object
This holds a text to be generated onto the page, this is used for all text to be generated onto a page whether it be a single header added directly to BtPage object or added to a [PageTable object](#pagetable-object). This object has properties like Font to display correctly onto the page.

##### PageTable object
This object takes a QList\<QStringList> to generate a table on a BtPage. For now this will auto-size it self to the content of the table, but this is subject to change in the future to be able to set in more detail all columns and rows sizes to better "style" the table so to speak.\
This also contains a QList\<[PageTableStyle](#pagetablestyle-object)> which is a **small** implementation on styling the table, read more below.
You define columns by adding a [PageTableColumn](#pagetablecolumn-object) to the PageTable::columnHeaders pointer. these contain the formatting for the column and the header text.

##### PageImage object
Draws an image on the page (aspiring MUC-Award winner! :) )\
Contaings the size of the imge and can resize when needed.

##### PageBreak object
This small little class does one thing, and one thing only....\
Signals the page renderer that there should be a page break before drawing the next object.

#### PageTableStyle object
Create this style as you wish( read 'can') and then add it to the PageTable object to style the tables output on screen.
#### PageTableColumn object
defines a column with attributes ina table. see [PageTable](#pagetable-object).

#### PlacingFlags
Places an object on the page or relational to another object according to the flags below.\
they can be concatinated by using the | operator.\
Example:
`page.placeOnPage(myChildObject, PlacingFlags::TOP | PlacingFlags::RIGHT, 0, -10)`\

Definition:

                enum struct PlacingFlags
                {
                   NOTUSED      = 0,
                   BELOW        = (1<<0), //00000000 00000000 00000000 00000001 (1)
                   ABOVE        = (1<<1), //00000000 00000000 00000000 00000010 (2)
                   RIGHTOF      = (1<<2), //00000000 00000000 00000000 00000100 (4)
                   LEFTOF       = (1<<3), //00000000 00000000 00000000 00001000 (8)
                   LEFT         = (1<<4), //00000000 00000000 00000000 00010000 (16)
                   RIGHT        = (1<<5), //00000000 00000000 00000000 00100000 (32)
                   TOP          = (1<<6), //00000000 00000000 00000000 01000000 (64)
                   BOTTOM       = (1<<7), //00000000 00000000 00000000 10000000 (128)
                   VCENTER      = (1<<8), //00000000 00000000 00000001 00000000 (256)
                   HCENTER      = (1<<9)  //00000000 00000000 00000010 00000000 (512)
                };

### Example:
This will create a page, add the recipe and Brewers name as well as Brewtargets logo to it and then render it to the printer.

```
   using namespace nBtPage;
   QPrinter *printer;
   BtPage page(printer);
   // adding the Recipe name as a title.
   PageText *recipeText = page.addChildObject(
      new PageText (
         &page,
         _parent->currentRecipe()->name(),
         QFont("Arial", 18, QFont::Bold)
      ),
      QPoint(0,0)
      );
   // adding Brewers name
   PageText *brewerText = page.addChildObject(
      new PageText (
         &page,
         _parent->currentRecipe()->brewer(),
         QFont("Arial", 10)
      ));
   page.placeRelationalToMM(brewerText, recipeText, PlacingFlags::BELOW, 2, 0);

   //Adding the Brewtarget logo.
   PageImage *img = page.addChildObject(
      new PageImage (
         &page,
         QPoint(),
         QImage(":/images/title.svg")
      ));
   img->setImageSizeMM(100, 20);
   page.placeOnPage(img, PlacingFlags::TOP | PlacingFlags::RIGHT);

   page.renderPage();

```

***Trivia:***\
_MUC Awards = Most Useless Comment Awards"_

### DIFF
--- a/dev-doc/BtPage.markdown
+++ b/dev-doc/BtPage.markdown
@@ -1,47 +1,47 @@
 # PrintPreviewDialog
->This dialog is an attempt to get the printing to be *cross-platform-WYSIWYG* in that it uses the QT standard way of generating the printout and preview.
+This dialog is an attempt to get the printing to be *cross-platform-WYSIWYG* in that it uses the QT standard way of generating the printout and preview.
 I also aimed to make a unified print/export dialog instead of having multiple menu items saying print-this and print-that.
 
-## nBtPage namespace
->I created this namespace to keep averything in one bubble of code and separate from the rest of Brewtarget. this was to minimize the risk of running into duplicate nameing.
+## BtPage namespace
+I created this namespace to keep averything in one bubble of code and separate from the rest of Brewtarget. this was to minimize the risk of running into duplicate nameing.
 This namespace includes all that is necessary for generating the printout and Html-export to file, while relying on Brewtargets RecipeFormatter- and BrewNoteFormatter-class to get the data for generating the pages. I'm not sure this is the way it should be in the future but will leave it there for now.
 
-### BtPage class
->This class is used to store all child elements/classes and placing them on the page. this class also contains the referens to the printer in use and is responible for calling all childrens ::render() method to get everthing printed onto the page.
+### Page class
+This class is used to store all child elements/classes and placing them on the page. this class also contains the referens to the printer in use and is responible for calling all childrens ::render() method to get everthing printed onto the page.
 This is genereated everytime you change your selection for printout or the prewview needs to refresh for any reason.
->#### Funtions:
->- **template \<class T> auto addChildObject(T \*obj, QPoint position = QPoint()) -> decltype(obj)**\
+#### Funtions:
+- **template \<class T> auto addChildObject(T \*obj, QPoint position = QPoint()) -> decltype(obj)**\
 You add content to the page by creating and adding PageChildObject-subclasses, see below. this returns a pointer to the added object if you need to make more setting to is, i.e. adding a style to a table
->- **renderPage()**\
+- **renderPage()**\
 Renders the page on the set printer/painter
->- **template \<class T, class S> void placeRelationalTo(T \*targetObj, S \*sourceObj, PlacingFlags place, int xOffset = 0 /\* pixels \*/, int yOffset = 0 /\* pixels \*/)**\
-Sets the postition of targetObj relational to the sourceObj according to the placingflags. you can also offset the postion using the Offset arguments.
->- **template \<class T, class S> void placeRelationalToMM(T \*targetObj, S \*sourceObj, PlacingFlags place, int xOffset = 0 /\* Millimeter \*/ , int yOffset = 0 /\* Millimeter \*/)**\
+- **template \<class T, class S> void placeRelationalTo(T \*targetObj, S \*sourceObj, RelationalPlacingFlags place, int xOffset = 0 /\* pixels \*/, int yOffset = 0 /\* pixels \*/)**\
+Sets the postition of targetObj relational to the sourceObj according to the RelationalPlacingFlags. you can also offset the postion using the Offset arguments.
+- **template \<class T, class S> void placeRelationalToMM(T \*targetObj, S \*sourceObj, RelationalPlacingFlags place, int xOffset = 0 /\* Millimeter \*/ , int yOffset = 0 /\* Millimeter \*/)**\
 Same as placeRelationalTo(..) but offset is in millimeters and this will calculate the distance in pixels according to the set printer and then call the original function with those values.
->- **template \<class T> void placeOnPage(T \*targetObj, PlacingFlags place, int xOffset = 0, int yOffset = 0)**\
-This will set the position on the supplied targetObj according to the PlacingFlags provided. this funtion is a convinient way of placing anyting to, for example, the TOP-RIGHT corner of the page or why not the center of the page. more on [PlacingFlags](#placingflags) below.
+- **template \<class T> void placeOnPage(T \*targetObj, FixedPlacingFlags place, int xOffset = 0, int yOffset = 0)**\
+This will set the position on the supplied targetObj according to the FixedPlacingFlags provided. this funtion is a convinient way of placing anyting to, for example, the TOP-RIGHT corner of the page or why not the center of the page. more on [PlacingFlags](#placingflags) below.
 ### PageChildObject
-> This is a abstract / Interface class for all child object to put on the page. It holds some basic informations about the object and makes sure that all derived classes implement a few methods, i.e. a render method, because all children need to know how to draw it self onto a QPainter and ultimatly onto a page.\
+ This is a abstract / Interface class for all child object to put on the page. It holds some basic informations about the object and makes sure that all derived classes implement a few methods, i.e. a render method, because all children need to know how to draw it self onto a QPainter and ultimatly onto a page.\
 This class also contains information on the boundingbox and position that is the same for all derived classes.
 
 #### Derived Classes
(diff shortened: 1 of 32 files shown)


matty0ung:
A couple of initial thoughts:

- I don't like `nBtPage` as a namespace name.  Why not have `BtPage` as the namespace and then call the BtPage class something else but put it inside the namespace?
- It feels like we're doing lots of painful string concatenation to create HTML docs.  Could we use something such as https://github.com/robertknight/qt-mustache to make our lives easier?

matty0ung:
Or if it really doesn't make sense for the class to be called anything other than BtPage, how about BtPageUtils or BtPageWrapper or some such for the namespace name?

mattiasmaahl:
I agree with you fully, I'll rename the namespace to BtPage. It actually was named theat when I created the namespace but then i couldn't think of a better name for the class, but that can be called just Page, that way it will be consisten with the rest of the "child" classes as well. 

Regarding the HTML creation I agree, that must be dona a better way, but I only used what was already there for now thinking that would be a later problem to refactor, this pkll request is getting somewhat big already and I'm not done yet. I'll look into the link you supplied maybe I can make something in this PR if it's not to much work.

mattiasmaahl:
Oh, and thanks for the input! 😊

matty0ung:
Some more random thoughts...

A couple of questions on `enum struct PlacingFlags`.  From the comment, it seems like it's two groups of flags, one for `placeRelationalTo()` and another for `placeOnPage()`.  Should it be two enums therefore, or is there a function that takes both combined?  Also, is there any potential mileage in reusing `enum Qt::AlignmentFlag`?

FYI also I think you mean `needPageBreak` rather than `needPageBrake` etc.  (You have PageBreak.h correct.)

More importantly, I'd avoid having variable names or #defines that start with underscore (eg your include guards and _children) as such names are technically reserved in a lot of circumstances (see https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier).

mattiasmaahl:
You make good points! Thank you for your time, much appreciated!
Yes, maybe PlacingFlags should be two separate enums? Let me thinknabout that.
Also, yes maybe Qt::AlignmentFlags would be a better way to go, but this solution was a sort of practice/learning experience with using flags and the | operator. But I'm not married to the solution it can be either way.

Regarding needPageBrake/needPageBreak you are correct, that's my misstake. Call it code blindness! 😳😊

The #define-underscore issue I did not know about, I just copied the leadtext/copyright from a random file and the kept with how it was written there, never giving much thought. I'll correct it monday. The _children private variable is a habbit from my python programming where you signal psuedo-private by naming variables with leading underscore. I'll fix it monday.

mikfire:
That's why you will see m_ used in several places. After a while, it gets quite natural and it is slightly annoying when I'm in older parts of the code.

I will review this. Work has been nuts lately, so I simply haven't had the spare cycles.

matty0ung:
You can tell I used to be a Java coder because I tend to write this-> for accessing member variables. :)  It is less concise than a prefix on the variable name, but it has the advantage that you can use it for (non-static) member functions too.  Still, either approach works.

mattiasmaahl:
So I went over the code fixing the above suggestions. I'll start working on the Inventory printout next.
There is still the Styling to be done.

mattiasmaahl:
Yeah, so, I have a few places where I have used functions in the QT library that was introduced after 5.9 version.
I need to go through them and fix it. sorry if I'm spamming commits.
If you want I can retract the PR and fix it in private?

matty0ung:
Give us a shout when you're happy with it and we'll take another look.  I can't remember what we decided the oldest version of Qt we needed to support was.  (I've got Qt 5.12 on Ubuntu 20.04.)

mattiasmaahl:
I believe it was decided to support version 5.9 for now, due to some linux distro.

I'll take the PR Back and rework a bit more, I need to clean up the previous version of printout/export that I'm replacing also, and I want to rework the formatters a bit as well. I'll be back with a new PR when that is done.

Issue #601, opened 2021-06-30
Inventory error?! Unknown signal. Trying to add inventory to yeasts.


mattiasmaahl:
I'm trying to add inventory to the Yeasts, one of the included ones (Danstar - Nottingham).
But Brewtarget crashes abruptly with the following error:
![bild](https://user-images.githubusercontent.com/1370239/123970551-df1bf900-d9c1-11eb-8801-a633e7ecb8c0.png)

 
![bild](https://user-images.githubusercontent.com/1370239/123970710-07a3f300-d9c2-11eb-97d6-1dfda1480634.png)

can anyone shed some light on this as I don't fully follow the redo/undo methods?

mattiasmaahl:
So far I'm having problems with Yeast, Hops and Misc. all with the above error.
Fermentables was OK to add to inventory for some reason.

This could be because of my database as well, have to do more testing tomorrow.

matty0ung:
I can reproduce this.  Will have a look.

matty0ung:
OK, I have a fix for this.  Will backport it to Brewtarget and submit in the next day or so.  (Need a couple of tweaks for the backport as `NamedEntityWithInventory` doesn't yet exist in BT :o>)

mattiasmaahl:
Nice, I'll continue my work with inventory printout using fermentables for now then awaiting your fix. 😊

matty0ung:
The crash was caused by an invalid property name in YeastTabelModel.cpp (and equivalent places for other things with inventory).  I've changed the hard-coded string property names for symbolic ones (so we should get compiler errors in the future for bad property names), as has already been done in most other parts of the code.  I also pulled out the common inventory functions from Yeast, Hop, etc to an intermediate parent class (NamedEntityWithInventory).  So the crashing seems to stop now...

However, there is still an issue that we're not always setting the inventory ID, which means inventory amounts aren't getting stored.  There may be some small fix in Brewtarget that addresses this, and I'm just not instantly seeing it.  Otherwise, when I bring the new DB layer back, there's new code (in a new Inventory class) that does all the right things.

mattiasmaahl:
So the error has been solved. I'll close this issue.
Solved in #602

PR #602, opened 2021-07-01
Partial fix for inventory bug

matty0ung:
This stops the crashing reported in https://github.com/Brewtarget/brewtarget/issues/601, but doesn't fix the problem that not everything has an inventory ID.  (This is fixed in Brewken because, when I rewrote the DB layer, I added some logic that automatically creates an inventory ID for anything that doesn't have one.  That fix will get backported soon, I hope, when I backport the new database layer.)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,6 +4,9 @@
 # See also ../CMakeLists.txt
 #
 
+# Uncomment the next line for verbose build output
+#set(CMAKE_VERBOSE_MAKEFILE ON)
+
 # The current directory in the original source.
 INCLUDE_DIRECTORIES( ${CMAKE_CURRENT_SOURCE_DIR} )
 # The current directory in the output build directory.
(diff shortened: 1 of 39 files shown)


mattiasmaahl:
Looks good, also compiles on my machine, although I had to install Boost libraries.
So we have to modify the checks/actions to accommodate this new library.
that is a small issue though.

mattiasmaahl (approved):
Looking good, I'll approve these changes in hope that the rest of the fix will come with the DB-backport at a later date. 😊
The Boost Library has to be added to the Actions but I'll fix that, I think.

matty0ung:
Thanks.  You could comment out the Boost include if it came to it.  It's just for debugging.  But I think there's a good chance we will end up using another bit of Boost for BeerJSON, so it's worth having it in the build.

mattiasmaahl:
That's what I thought. I have a fix on its way. Creating PR as we speak.

Issue #606, opened 2021-07-03
Editing mash crashes the program


Drevi:
Modifying anything in the mash tab causes the following crash:
![image](https://user-images.githubusercontent.com/33293765/124348682-b7937f00-dbc1-11eb-9bd0-2b1cd78d01d5.png)

It doesn't matter if it's amount, time, temp or if you edit by hand or the mash wizard/designer. Tried both with my own recipes on an imported db from 2.3, and on an empty fresh 2.4 db. 

The crash log:
> [05:41:52.268] INFO : QString Database::findRecipeFromForeignKey(TableSchema*, const NamedEntity*) couldn't find a key for "", in 1085

matty0ung:
Ah, hmm, yes, that happens for me too.  I'll see if there's a quick fix.  (If not, the new database code I'll shortly backport from Brewken to Brewtarget does not have this bug.  So, one way or another it will get resolved.)

matty0ung:
OK, I know what's going on here.  There's a bug in Database::modifyEntry() where, in certain circumstances, it is calling the wrong version getParentRecipe().  We need to do some downcasting in one of the functions to take account of the fact that MashStep and BrewNote have their own special versions of getParentRecipe()  I'll work on a fix.

matty0ung:
I've made fix.  It's a bit ugly, but it works and, in any case, will get changed when I do the new DB layer.  (I'm actually part-way through porting forward @mikfire's Recipe versioning code to Brewken as a prelude to being able to backport the new DB layer to Brewtarget.  So this is hopefully within the foreseeable future.)

Drevi:
Thanks, working flawlessly with this fix.

PR #607, opened 2021-07-03
Wrong version of getParentRecipe called for MashStep and BrewNote from recipe versioning function

matty0ung:
Fix for https://github.com/Brewtarget/brewtarget/issues/606


mattiasmaahl (approved):
If it works, it works!

PR #617, opened 2021-07-26
New Database Layer

matty0ung:
This is the new database layer, which separates out:
- mapping objects to/from database storage
- dealing with the specifics of an individual database (credentials, backups, different ways of doing things in SQLite/PostgreSQL, etc)

It also implements export to BeerXML using the same new data structures etc we created to improve reading from BeerXML.

A few other things have come along for the ride because they were minor tidy-ups I'd implemented in Brewken and it would have been more effort to exclude them than to include them.

All database stuff is now in the database subfolder of src.  Key files:
- `Database.cpp, Database.h` handle abstraction of SQLite / PostgreSQL (as before) but no longer do any object mapping
- `DatabaseSchemaHelper.cpp, DatabaseSchemaHelper.h` handle schema upgrades and migrations, plus default data population (as before).  I've disentangled this from the old *Schema* files that are no longer used.  In particular schema changes are now hard-coded.  (If we've got the SQL that, say, upgrades from v7 to v8 of the DB, we don't want this to be tied to "current" mapping as you still want to be able to do that exact same v7 to v8 step (and v8 to v9 etc etc) at any later date, including when the "current" schema has evolved and changed from whatever v7 was.)  I have also moved "default data" to BeerXML and stopped using the bt_ tables.  (Now that we decided not to try to edit the user's existing data, we only needed the bt_ tables for duplicate detection - ie not to import the default Recipe/Style/Hop/etc objects that are already in the user's database.  But we have such logic already in the BeerXML import code.  So we can just reuse that and it will automatically pull only "new" objects out of `data/DefaultData.xml`.  Of course, once we support BeerJSON, we'll switch to that.)
- `ObjectStore.cpp, ObjectStore.h` Does all the SQL to create/read/update/delete objects in the database.  Normally you won't access this directly, but instead via templated functions in the  `ObjectStoreWrapper` namespace
- `ObjectStoreTyped.h, ObjectStoreWrapper.h` add templated wrappers around ObjectStore to save callers doing lots of tedious casting
- `ObjectStoreTyped.cpp` contains all the data required to map objects to and from database tables
- `DbTransaction.cpp, DbTransaction.h` just provide an RAII wrapper around DB transactions

A fair bit of class-specific logic that used to be inside the Database class is now inside the individual subclasses of `NamedEntity` (in the `src/model` directory).  Eg, more of the Recipe versioning logic now lives inside the Recipe class.  I've tidied up a few things with helper functions so that we have less copy-and-paste code on setter member functions.

Generally, you'll see that:
- Dealing with storing and retrieving objects is now much more C++ than SQL.  Eg, if you want "all Recipes that match such-and-such condition" then that condition is going to be a lambda that you pass to `ObjectStoreWrapper::findAllMatching<Recipe>()` rather than a `WHERE` clause that you tack onto a SQL statement in a new `Database` member function.  Similarly, default data for a newly-created object is always in that object's constructor and not in the database definition.
- The mappings from objects to database tables are more concise and use fewer global constants.
- We have better encapsulation as fewer classes need to be friends with each other, and classes in the `src/model` directory do not need to have any knowledge of SQL or XML.
- Inside the object model, we do more lazy-loading (eg a Recipe object holds the IDs of the Hops it uses and then turns those IDs into Hop objects if and when needed), but this should be largely invisible to the UI code.
- It should now be less work in future to add support for another SQL database (eg MariaDB) or even a NoSQL (eg CouchDB or Cassandra).

Additionally, I have:
- Tidied up logging a bit, including now logging a thread ID (in case we do any multi-threaded stuff in future) and ensuring that errors about logging itself are always logged to the console even if we have disabled other console logging
- Enhanced the GitHub actions for the Mac build so that you can see a stack trace if the automated tests crash. (This is what happens when you don't own a Mac!)
- Separated out persistent settings into a namespace called `PersistentSettings`
- Created a new class `BtStringConst` for wrapping compile-time-constant strings. (See class comment in `utils/BtStringConst.h` for part of the motivation for this but, amongst other things, this made the Windows and Mac builds happy.)
- Combined the above two steps to use proper constants for all access of persistent settings (which along the way uncovered a few minor bugs where, eg, we were using one key to store a value and another one to attempt to read it)

This all turned out to be a bit more work than I imagined back in January, but I'm happy with the result.  There's scope to refine things a bit further (eg I think some or all of ObjectStoreTyped could be merged into ObjectStoreWrapper) but I don't see the need to do another mega-change like this one.

As ever, questions and comments most welcome.


matty0ung:
PS: See https://github.com/Brewtarget/brewtarget/pull/617/commits/94cfd7dbf470ee79400112f199c3962044f3ea5b for a motivation for storing default data in a non-binary format. :)

mikfire:
Are the boost backtrace libraries really required? cmake is really not happy with them, on either ubuntu 20.04.2 or gentoo. 

I am not even certain I can get them installed on gentoo, without extraordinary effort. Unless these libraries provide value greater than that effort, I would prefer not to invoke them.

matty0ung:
Hmm, that's odd.  Ubuntu 20.04.2 is what I run and I don't remember having to do anything special to get it to work.  (The Mac build was a bit of another story.)  But it's small work to disable it if you want.  The backtrace is just there for diagnostics.  I'll do a patch.

mikfire:
Thanks. I could get it working in an Ubuntu VM, but not on my gentoo host systems. It is a little embarrassing, actually. I am not used to being able to get something working on Ubuntu but not gentoo.

I am hesitant to approve this, to be honest. I like the approach. Database.cpp is blessedly short and does what I have wanted it to do for a long time.

There's a lot that needs to be fixed, and it is exactly what I expect when we go playing with the database layers. I am also concerned some of what I am seeing is not repeatable. I had one problem that just mysteriously went away, and I have no idea why.

A simple complexity issue. There four entries when trying to select the database type on the options screen. I see two entries for SQLite and two for Postgresql. That is likely something isolated to the options screen and should be a simple fix.

A medium complexity issue. Change the style on any recipe and then go look at the Style tree. If you are seeing what I see, there are now two of that style listed. If you restart brewtarget, the duplicate goes away. This is likely to be an order of operations problem -- the tree is signaled to update before the "visible" flag is set to false. I am not fond of chasing these problems.

Two large complexity issues. postgresql does not seem to be working. Automatic recipe snapshots are dumping core.

So do we merge into develop, knowing we are breaking things? We can at least open proper tickets, and people can help resolve them. Are there other options?

matty0ung:
Thanks @mikfire.  Let me see if I can fix the Options screen thing.  Should be simple, as you say.  I'll then see if I can work out the signals order thing on styles.

For the snapshot core dump, do you have a stack trace?  (I wanted to talk about automatic snapshots, as after I'd jumped through a few hoops to get it to detect any change to any element of a recipe, I started wondering whether it was really helpful to have a new version of the recipe every time every single field was changed.  I wondered if it would be more helpful to say "After a Recipe has been brewed, the next change creates a new version.  No further new versions (of the recipe) are created automatically until after its next brew day."  What do you think?)

Similarly, for PostgreSQL, are you seeing any useful diagnostics?

mikfire:
It isn't just on styles. So far, my testing indicates it effects everything. However, odds are good that fixing for one will fix it for all. I think it has to be a signal issue, since restarting brewtarget removes the duplicate entry. The tree is either being signaled before display is set false, or a filter isn't being used that should be.

I am testing on a different VM now, and things are inconsistent. 

The automatic snapshot created 30 snapshots in response to me adding one ingredient. My guess for that would be a snapshot was taken for every thing that got copied from the parent to the child. Previously, it just dumped core. They could be the same issue, and I just got lucky I didn't run off the stack this time.

And postgresql is not working. The upgrade fails with a
```
[15:20:29.865] (1did888j0g) ERROR : bool {anonymous}::executeSqlQueries(QSqlQuery&, const QVector<{anonymous}::QueryAndParameters>&) Error executing database upgrade/set-up query  "ALTER TABLE recipe ADD COLUMN ancestor_id INTEGER REFERENCES recipe(id)" :  "ERROR:  syntax error at end of input\nLINE 1: EXECUTE \n                ^\n(42601) QPSQL: Unable to create query"  [database/DatabaseSchemaHelper.cpp:92]
```

I am thinking that may be the result of how a QString is being built. I can keep poking at that, as I am the only person who cares about using postgresql.

mikfire:
If I manually do the upgrade from version 9 -> 10 (add two columns, set some default values and then update the db version to 10), the postgresql access works.

There is definitely something off in the upgrade. I might have some older databases laying about. Let me see if this is just a 9 -> 10 problem, or if it is a general upgrade problem.

To answer your other question. Yes, that was the way I had designed the automatic snapshots. Once a recipe was brewed, it was "soft locked" and the first change would spawn a child. Any subsequent change would not spawn a child until it was brewed again.

There may have been a difference between what I designed and what I delivered.

matty0ung:
Ah, OK, good.  Couple of small fixes above.  The versioning thing is my misunderstanding of the functionality.  Should be able to get it closer to what you had.  (Pulling this logic out of the Database layer involved quite a lot of change, though it also made me clean up all the setters, so it had good side-effects!)  There will doubtless be still one or two bugs, but hopefully then more debuggable.

The upgrade stuff will be easy to fix, if that's the only place you're getting errors.  I was doing the PostgreSQL-specific stuff without having an instance of PostgreSQL to test on, so I'm hoping it's mostly just typos etc on my part.  (Differences in SQL between SQLite and PostgreSQL are mostly OK, just a bit tedious when they need things in a different order.)

matty0ung:
With the upgrade error, it's probably staring me in the face, but can you say why PostgreSQL is giving SYNTAX ERROR for
```
ALTER TABLE recipe
ADD COLUMN ancestor_id INTEGER
REFERENCES recipe(id)
```
The answer from Evan Carroll at https://dba.stackexchange.com/questions/132029/how-to-add-a-column-with-a-foreign-key-constraint-to-a-table-that-already-exists suggested this was an OK way to add a new column with FK constraint to PostgreSQL.

mikfire:
I found something that claims to be a version 7 database, so I am using that. It is my standard mess of a database, but I think I can make it look more normal.

The first issue is in DatabaseSchemaHelper when creating any auto-incrementing primary key. In SQLite, it has to be "id INTEGER PRIMARY KEY". In postgresql, it has to be "id SERIAL PRIMARY KEY". Including the "INTEGER" is causing postgresql all sorts of heartburn. My preference would to rename Database::getDbNativeIntPrimaryKeyModifier() to be something like "getDbNativePrimaryKeyDeclaration()"
return the entire string instead of building it in place.

The second issue I am seeing is that postgresql uses "true/false" for boolean values. SQLite uses "1/0". Trying to use "1/0" in postgresql will cause an error. I see 1 and 0 hard coded in several places in DatabaseSchemaHelper. The reason Database::dbBoolean exists is to handle this problem, and I still find Database::dbTrue() and Database::dbFalse() to be useful shortcuts. I don't know if that problem is more wide spread or not.

I think the problem you are looking at isn't a syntax error in the SQL. As far as I can determine, postgresql does not like it when you prepare a statement that has no bind values. I am uncertain if it doesn't like the fact nothing was bound (ie, sql.prepare, no sql.bind, sql.exec) or it doesn't like to prepare a statement with no ? in it. Regardless, it seems to think this is a syntax error. The fix I used was to modify DatabaseSchemaHelper::executeSqlQueries() to prepare, bind and exec only if there are values. Otherwise, it just execs the string. That seems to resolve this issue.

There are several places in DatabaseSchemaHelper::migrate_to_8 where it says "ALTER TABLE [thingy] ADD COLUMN inventory_id REFERENCES" which needs to have an "INTEGER" between the id and REFERENCES.

I am a little concerned in Database::getDbNativeIntPrimaryKeyModifier (and it's friends) take a DbType parameter, but don't use it. I haven't gotten this deep yet, but that has a real potential to cause problems when copying one database to another. You may be handling it different, but it did catch my eye.

This has gotten long. I still haven't successfully upgraded from 7 -> 8 yet, but I will stop on this thread.

mikfire:
The migrate_to_8 is painful. I question how I even figured it all out.

migrate_to_8 is failing under the new code with this error:
```
[11:33:12.773] (1dobw33pmo) ERROR : bool {anonymous}::executeSqlQueries(QSqlQuery&, const QVector<{anonymous}::QueryAndParameters>&) Error executing database upgrade/set-up query  "
INSERT INTO fermentable_in_inventory (fermentable_id) VALUES ( 
   SELECT id FROM fermentable WHERE NOT EXISTS ( 
      SELECT fermentable_children.id FROM fermentable_children 
        WHERE fermentable_children.child_id = fermentable.id ) 
   AND NOT EXISTS ( 
      SELECT fermentable_in_inventory.id FROM fermentable_in_inventory 
        WHERE fermentable_in_inventory.fermentable_id = fermentable.id) 
   )" :  
"ERROR:  syntax error at or near \"SELECT\"\nLINE 1: ...ermentable_in_inventory (fermentable_id) VALUES ( SELECT id ...\n                                                             ^\n(42601) QPSQL: Unable to create query"  [database/DatabaseSchemaHelper.cpp:98]
```

I believe the problem here is that I don't track any inventory. I don't actually use the inventory subsystem at all. So the inner query returns empty, and trying to `INSERT INTO fermentable_in_inventory (fermentable_id) VALUES ()` throws an error. The while loop in the original code got around this problem by not doing anything if nothing returned from the select. I have no suggestions on how to fix this one.

I am also somewhat blocked on more testing until I can figure out how to get around this.

matty0ung:
Hmm, good question.  I can't find a way to do it in SQL, but I might have an idea about how to handle this...

matty0ung:
Going back to your prior points:

Thanks for info on the primary key in PostgreSQL.  Should be an easy fix.

On booleans, AIUI, it should be possible to just use true/false everywhere in SQL and the drivers will figure out to translate that into 0/1 for SQLite.  So, if we've got hard-coded 0s and 1s, I just need to change them.

On the prepare stuff, I would never have guessed about that limitation from the Qt documentation.  Thanks for working it out!

On the getDbNativeIntPrimaryKeyModifier() DbType parameter, you're right, it shouldn't be there.  My approach to copying from one type of DB to another was to have a separate instance of Database for the from and the to databases (see `DatabaseSchemaHelper::copyDatabase()`).

mikfire:
Matt, I really hope I am not annoying you or otherwise antagonizing you. I am simply running through my standard tests and trying to doc what happened. If I am getting on your nerves or if you need a break, just let me know and I will hold off on more.

I am now trying to copy SQLite->postgresql. It isn't quite as gruesome. We already know most of the fixes. A few small changes and this seems functional. I will, of course, try it the other way. But usually getting it working in one direction almost always works in the other.

Before the usual laundry list of issues, why are the database tables being created without default values? I am not saying it is a good or bad thing. I just want to understand the reasoning.

Also, why is copyDatabase() in DatabaseSchemaHelper? Again, I am not arguing that it shouldn't be. My first thought is that it goes in Database. 

The same "INTEGER" problem with the primary keys was found. I think that can be fixed by fixing ObjectStore::createTableWithoutForeignKeys to not use getDatabaseNativeTypeName when firstFieldOutput is false.

ObjectStore::createTableWithoutForeignKeys also has the problem of preparing with no binds.

`ERROR:  parameter $7 of type integer cannot be coerced to the expected type boolean`. The boolean representation in the two databases is a massive pain. I don't claim to know what the QVariant is doing here, but I can promise it is not doing what we need. That is why Database::dbBoolean() exists -- there is no way to get it done. DatabaseSchemaHelper::copyDatabase() probably needs something like this
```c++
if ( field.type() == QVariant::Bool ) {
    bindValue = newDatabase.dbBoolean(bindValue.toBool());
}
```

As a rule, I think in the error case we need to first tell the user something went wrong. We also probably should consider not changing the dbType in the config file, but I think that has always worked that way?

mikfire:
> On the getDbNativeIntPrimaryKeyModifier() DbType parameter, you're right, it shouldn't be there. My approach to copying from one type of DB to another was to have a separate instance of Database for the from and the to databases (see `DatabaseSchemaHelper::copyDatabase()`).

You know, I saw that once I was digging into that part of the code. Perfectly acceptable solution. Could we could leverage that to just switch to the new database once it's copied, instead of asking the user to restart?

matty0ung:
> Matt, I really hope I am not annoying you or otherwise antagonizing you. I am simply running through my standard tests and trying to doc what happened. If I am getting on your nerves or if you need a break, just let me know and I will hold off on more.
No, it's all good stuff if we're fixing bugs and making the code clearer. TBH on a change this size it would be surprising if we weren't finding issues. Keep throwing things my way and I'll work through it all.

matty0ung:
> why are the database tables being created without default values?
It's a fair question. My thinking was that we will never use DB default values because defaults are set in the object constructors of Recipe/Hop/Fermentable/etc.

matty0ung:
> Why is `copyDatabase()` in DatabaseSchemaHelper [rather than Database]?

My thinking here was that Database should not have any knowledge of the schema or the object stores.  I did toy with the idea of doing copyDatabase in some schema-agnostic way (ask the database for a list of all its tables, and for each table ask for a list of all its columns etc) but I decided it would be simpler and safer to use ObjectStore (because eg we don't want to pick up special tables the DB uses for sequences).

That said, I think you're right that `copyDatabase()` isn't sitting in the best place right now.  I'll move it into the ObjectStore code.

matty0ung:
> ObjectStore::createTableWithoutForeignKeys also has the problem of preparing with no binds

This was annoying, so I created new class `BtSqlQuery` (to replace `QSqlQuery`) that should behave the same on PostgreSQL and SQLite (ie won't mind you preparing a statement that doesn't have any bind placeholders).

matty0ung:
> The same "INTEGER" problem with the primary keys was found. I think that can be fixed by fixing ObjectStore::createTableWithoutForeignKeys to not use getDatabaseNativeTypeName when firstFieldOutput is false.

This should now be fixed (I hope!)

matty0ung:
> `ERROR: parameter $7 of type integer cannot be coerced to the expected type boolean.` The boolean representation in the two databases is a massive pain. I don't claim to know what the QVariant is doing here, but I can promise it is not doing what we need.

Hmm, yes, I see what you mean.  That's annoying because QVariant knows how to coerce an int to a bool (per doco for `QVariant::toBool()`) but I guess it must just be passing the raw int to PostgreSQL and it's the DB that is complaining about it not being a bool.

I think I may be able to take another approach here.  We don't actually necessarily need to read from the old database as we already have all the data cached in memory in the ObjectStore instances.  Let me see if I can get it to write everything out to new DB without having to mess the code around too much...

matty0ung:
> Could we leverage that to just switch to the new database once it's copied, instead of asking the user to restart?

Yes, I think we could.  It's probably not a huge code change, but would need some careful testing to make sure none of the code was still trying to do something in the old DB after the switchover.  Would you be OK for us to do that as a separate enhancement ticket?

matty0ung:
OK, I _think_ I've now, one way or another, addressed all the points in the comments.  Please let me know if I've missed anything or if you find any other issues.

mikfire:
I am going to try something new and different and only do one of these.

# ObjectStore.h

This may be my system, but I keep having to add a
```c++
#include <optional>
```
in ObjectStore.h

# Copying from SQLite to Postgresql

## repopulateChildrenOnNextStart

```
[10:41:17.030] (1dnxxf1v9c) ERROR : bool {anonymous}::executeSqlQueries(BtSqlQuery&, const QVector<{anonymous}::QueryAndParameters>&)
   Error executing database upgrade/set-up query
      "INSERT INTO settings (repopulatechildrenonnextstart, version) VALUES (?, ?)" :
   "ERROR:  parameter $1 of type boolean cannot be coerced to the expected type integer"
```

Did you know that, for reasons that escape me, the repopulateChildrenOnNextStart column was implemented as an integer and not a boolean?

### suggested fix
Probably can just change
```c++
{QString("INSERT INTO settings (repopulatechildrenonnextstart, version) VALUES (?, ?)"), {QVariant(true), QVariant(dbVersion)}}
```
to
```c++
{QString("INSERT INTO settings (repopulatechildrenonnextstart, version) VALUES (?, ?)"), {QVariant(1), QVariant(dbVersion)}}
```

in DatabaseSchemaHelper:754

## insertObjectInDb

Once I fix the previous issue, I am seeing this
```
[11:26:06.691] (1dvxtqy7b4) ERROR : int ObjectStore::impl::insertObjectInDb(QSqlDatabase&, const QObject&, bool)
   Error executing database query  "INSERT INTO brewnote ..." :
   "ERROR:  insert or update on table \"brewnote\" violates foreign key constraint \"brewnote_recipe_id_fkey\"
     DETAIL:  Key (recipe_id)=(31) is not present in table \"recipe\".\n(23503) QPSQL: Unable to create query"  [database/ObjectStore.cpp:747]
```

My databases run under normal users granted the owner role by the superuser and does not have permission to execute these commands. As you mention in the comments, the other solution requires the keys to be created in a particular way. My quick testing indicates that our keys are not so created.
```
brew001=*> set constraints brewnote_recipe_id_fkey deferred;
ERROR:  constraint "brewnote_recipe_id_fkey" is not deferrable
```

We could maybe update the create commands, I don't think that will do us any good. Any existing postgresql database (of which I may be the only user) won't have those definitions, and any process relying on the keys being properly defined simply won't work. To add insult to injury, the `SET CONSTRAINTS` method has to be done inside the transaction for postgresql but the `PRAGMA` must be done outside the transaction for SQLite.


### suggested fix
Given that we allow the user to store the credentials in a file in plain text, I am really uncomfortable with requiring them to provide a user with the necessary juice. I am not even certain I know the password for my superuser.

One thing I can think to do is to prompt for the superuser privileges before we start the copy and bind to postgresql with that ID instead of the stored one? That way, at least, the password is only ever memory resident? That could simplify setting up the postgresql. Right now, it requires a person to create the database manually. If we had superuser credentials, we could do that as well.

This worked in previous code because we did things in a specific order and made sure we did it in that order. It isn't pretty, but this would work if we deferred the brewnote objects until after the recipes. I know because I deleted all the brewnotes in my test db and this error went away.

## Brewtarget tooltip
If you haven't done this, you may not realize there is a pop-up when you hover over the Brewtarget image above the trees. It is supposed to show the database type you are connected to, and some useful information with respect to that database.

Even though I am connected to a sqlite database, it is showing information for a postgresql database. This has confused me several times. This may not be a problem with this PR. I will check older code.

## coredump on db transfer
This is causing some fun side effects.

```
[13:08:15.578] (1dvxtqy7b4) ERROR : int ObjectStore::impl::insertObjectInDb(QSqlDatabase&, const QObject&, bool) After writing Equipment # 26 to database, primary key was changed to 1  [database/ObjectStore.cpp:782]
[13:08:15.578] (1dvxtqy7b4) ERROR : ASSERT: "false" in file /home/mik/brewtarget/matty0ung/src/database/ObjectStore.cpp, line 785  [database/ObjectStore.cpp:785]

Thread 1 "brewtarget" received signal SIGABRT, Aborted.
0x00007ffff5e8029e in raise () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff5e8029e in raise () from /lib64/libc.so.6
#1  0x00007ffff5e6a536 in abort () from /lib64/libc.so.6
#2  0x00007ffff63f7b17 in QMessageLogger::fatal(char const*, ...) const () from /usr/lib64/libQt5Core.so.5
#3  0x00007ffff63f6e3d in qt_assert(char const*, char const*, int) () from /usr/lib64/libQt5Core.so.5
#4  0x000055555569c171 in ObjectStore::impl::insertObjectInDb (this=0x555555b92240, connection=..., object=..., writePrimaryKey=true)
    at /home/mik/brewtarget/matty0ung/src/database/ObjectStore.cpp:785
#5  0x000055555569947f in ObjectStore::writeAllToNewDb (this=0x555555b34800 <(anonymous namespace)::ostSingleton<Equipment>>, connectionNew=...)
    at /home/mik/brewtarget/matty0ung/src/database/ObjectStore.cpp:1500
#6  0x00005555556a2a47 in WriteAllObjectStoresToNewDb (newDatabase=..., connectionNew=...) at /home/mik/brewtarget/matty0ung/src/database/ObjectStoreTyped.cpp:804
#7  0x000055555568c694 in DatabaseSchemaHelper::copyToNewDatabase (newDatabase=..., connectionNew=...) at /home/mik/brewtarget/matty0ung/src/database/DatabaseSchemaHelper.cpp:852
#8  0x0000555555681540 in Database::convertDatabase (this=0x555555b33890 <Database::instance(Database::DbType)::dbSingleton_SQLite>, Hostname=..., DbName=..., Username=..., Password=...,
    Portnum=5432, newType=Database::PGSQL) at /home/mik/brewtarget/matty0ung/src/database/Database.cpp:907
#9  0x00005555558281c6 in OptionDialog::transferDatabase (this=0x555556225e80) at /home/mik/brewtarget/matty0ung/src/OptionDialog.cpp:881
```

I have 35 equipment profiles total in my test database, so the \#26 is odd. I've also see this a few times in different places, I've also not seen it a few times which generates the next core dump too.

The net result is that no data gets transferred, as this dump or the next rolls the transaction back.

## coredump even when no transfer
If I just switch database types and DO NOT transfer the data, I get this core:

```
[13:12:25.406] (1dvxtqy7b4) ERROR : bool Database::impl::loadSQLite(Database&) 
Error executing database query  "
SELECT sqlite_version() AS version;" :  
"ERROR:  function sqlite_version() does not exist\n\
LINE 1: SELECT sqlite_version() AS version;\n               ^\nHINT:  No function matches the given name and argument types. You might need to add explicit type casts.\n(42883) 
QPSQL: Unable to create query"  [database/Database.cpp:273]
WARNING:  there is no transaction in progress
[Thread 0x7ffff4ac5640 (LWP 26030) exited]

Thread 1 "brewtarget" received signal SIGSEGV, Segmentation fault.
0x00007ffff6430e07 in QReadWriteLock::lockForRead() () from /usr/lib64/libQt5Core.so.5
(gdb) bt
#0  0x00007ffff6430e07 in QReadWriteLock::lockForRead() () from /usr/lib64/libQt5Core.so.5
#1  0x00007ffff7f3035a in QSqlDatabase::connectionNames() () from /usr/lib64/libQt5Sql.so.5
#2  0x00005555556803c6 in Database::unload (this=0x555555b338a0 <Database::instance(Database::DbType)::dbSingleton_PostgresSQL>) at /home/mik/brewtarget/matty0ung/src/database/Database.cpp:731
#3  0x000055555567ebc8 in Database::~Database (this=0x555555b338a0 <Database::instance(Database::DbType)::dbSingleton_PostgresSQL>, __in_chrg=<optimized out>)
    at /home/mik/brewtarget/matty0ung/src/database/Database.cpp:515
#4  0x00007ffff5e82a33 in ?? () from /lib64/libc.so.6
#5  0x00007ffff5e82bda in exit () from /lib64/libc.so.6
#6  0x00007ffff5e6b804 in __libc_start_main () from /lib64/libc.so.6
#7  0x00005555555bb7ca in _start ()
```
### suggested fixes
This looks like a query is being run against the new database but still using the old databases information. I haven't a suggested fix just yet.


# upgrading existing postgresql

I don't see I can make much headway against copying the databases, so I will now try to upgrade some things. Remember, this is a mostly version 7 upgrade. One hopes the hard part is 7 -> 8, but lets see.

## booleans
Yeah. This again. It's a different error, but it's the same root problem.

```
2021-08-23 13:22:06.397 EDT [29844] STATEMENT:  PREPARE qpsqlpstmt_1 AS
  CREATE TABLE tmpbrewnote (
    id            SERIAL PRIMARY KEY,
    abv DOUBLE    PRECISION DEFAULT 0,
    attenuation   DOUBLE PRECISION DEFAULT 1,
    boil_off      DOUBLE PRECISION DEFAULT 1,
    brewdate      DATE DEFAULT CURRENT_TIMESTAMP,
    brewhouse_eff DOUBLE PRECISION DEFAULT 70,
    deleted       BOOLEAN DEFAULT $1,
    display       BOOLEAN DEFAULT $2,
    ...
```

I am guessing some prepared values never got bound? I can see other hints of this problem elsewhere.

### suggested (partial) solution
Since we are getting rid of the default values everywhere else, it probably makes more sense to get rid of them here as well. I think this is not really fixing the problem but I think it at least delays the problem.

## where's the INTEGER
The commands for doing this:
```sql
ALTER TABLE fermentable ADD COLUMN inventory_id REFERENCES fermentable_in_inventory (id)
```
are throwing errors in postgresql.

### suggested fix
Modify lines 392, 447, 486, 525 to look like this for postgresql:

```sql
ALTER TABLE <table> ADD COLUMN inventory_id INTEGER REFERENCES fermentable_in_inventory (id)
```

We already did this in Database, although it might be a little cleaner as we've hardcoded INTEGER into the string, and we should likely be using the `db.getDbNativeTypeName<int>()` instead. We can maybe kick that problem down the road a little, as any database that doesn't use INTEGER won't need upgrading. Yeah, it's a dirty cheat, but I think we can reasonably make some excuses in this bit of the code.

## last one for today

Ok. Last one for the day.

```
[13:51:07.977] (1dkx4g4ygw) ERROR : bool {anonymous}::executeSqlQueries(BtSqlQuery&, const QVector<{anonymous}::QueryAndParameters>&)
   Error executing database upgrade/set-up query
      "INSERT INTO hop_in_inventory (hop_id) VALUES (
         SELECT id FROM hop WHERE NOT EXISTS (
            SELECT hop_children.id FROM hop_children WHERE hop_children.child_id = hop.id )
            AND NOT EXISTS (
               SELECT hop_in_inventory.id FROM hop_in_inventory WHERE hop_in_inventory.hop_id = hop.id
            )
       )" :
   "ERROR:  syntax error at or near \"SELECT\"\nLINE 1:
      INSERT INTO hop_in_inventory (hop_id) VALUES (
      SELECT id FRO...\n                                                       ^\n(42601)
      QPSQL: Unable to create query"  [database/DatabaseSchemaHelper.cpp:94]
```

Same error as yesterday, but now for hops.


### suggested fix
I am uncertain if you were just testing to see if this worked. If you were, then it did work and now we either need to copy/paste that a few more times (hops, miscs and yeasts) or...

If I might humbly suggest, could we reconsider this a little? The one set of queries spans about 320 lines. Doing the simple copy/paste will add about another 40 or so lines. I think 360 lines is way too long for one data structure. At least, I am not finding the readability of this mega-query to be any greater than the readability of a doing this with a few while loops and some reasonable string substitutions?

mikfire:
> > Could we leverage that to just switch to the new database once it's copied, instead of asking the user to restart?
> 
> Yes, I think we could. It's probably not a huge code change, but would need some careful testing to make sure none of the code was still trying to do something in the old DB after the switchover. Would you be OK for us to do that as a separate enhancement ticket?
> 
> > As a rule, I think in the error case we need to first tell the user something went wrong. We also probably should consider not changing the dbType in the config file, but I think that has always worked that way?
> 
> Agreed. Are you OK to bundle up in the same ticket as above? (I'm happy to do the work, I'm also keen to draw this particular ticket to a conclusion if we can get it to a point where it's not taking any steps backwards! smile )

Definitely no problem doing this as a separate pull request. It's a thing that's annoyed me for a long time, and I was hoping if I mentioned it you would have better ideas :smile:

matty0ung:
This is good stuff -- thanks Mik.

- **ObjectStore.h** - My bad, will add the missing include here (and in OptionDialog.cpp)
- **repopulateChildrenOnNextStart** - Should have spotted that as it even says int on the previous line.  Will do a fix
- **insertObjectInDb** (need for superuser perms) - Should we look at storing passwords securely rather than in plain text?  Seems we could use code from https://github.com/frankosterfeld/qtkeychain to do most of the heavy lifting to have the OS store a password in its keychain / credential store / etc.  WDYT?
- **Brewtarget tooltip** - I hadn't spotted this.  Looks like the code is in `RecipeFormatter::getLabelToolTip()` which is not the most obvious location for it!  Don't see any reference to SQLite there.  My instinct would be for whatever code constructs the tooltip to ask the `Database` class for a list of name:value credentials.  Should be small effort I think, so will include with my other fixes.
- **coredump on db transfer** - The assert is saying that the primary key on the object we just inserted is not what we think it is.  There should be an error logged immediately before the app exited which shows actual and expected values.  If you have that to hand, it would give us a good clue as to what's going on.  (Might perhaps be that the DB doesn't return a primary key when we specified one rather than having the DB generate one.)
- **coredump even when no transfer** - Took me a few minutes to work this one out. I think the bug is in `dbConnectionNameForThisThread` where we're saying "each thread has one DB connection" instead of "each thread has one DB connection per database".  Will do a fix.
- **upgrading existing postgresql**
  - **boolean** - Ach, I thought I was being very clever using QVariant for the default values, but obviously not clever enough!  Your solution of dropping the defaults is much better.  Will implement.
  - **where's the INTEGER** - Yes, having written `Database::getSqlToAddColumnAsForeignKey()` we should make sure to actually use it when we add a column!  Will fix.
  - **last one for today** - Oh, yes, I see we need a dummy run query for this too and the other places you mention.  Your suggestion about loop and substitute sounds the right way to go.  I'll do a fix.

mikfire:
>     * **coredump on db transfer** - The assert is saying that the primary key on the object we just inserted is not what we think it is.  There should be an error logged immediately before the app exited which shows actual and expected values.  If you have that to hand, it would give us a good clue as to what's going on.  (Might perhaps be that the DB doesn't return a primary key when we specified one rather than having the DB generate one.)
> 

If you look at the first two lines of the output, it is there. We inserted equipment # 26 (out of 35) and got back 1. I've seen this succeed with no error. I also remember another core dump where the code inserted # 8 and got back 1 but I do not remember what table.

I will run that test a few times and see if I can capture a bit more. One thing I do like about postgresql is it has great logs.

mikfire:
**insertObjectInDb** (need for superuser perms) - For now, I suggest we mark this as a known bug and I use the admin id and password to further my testing. I want time to do this better, but I don't think my idea of "better" belongs in the scope of this PR.

There should be a new tool, something like "Create new database" which would allow you to specify which type you wanted to create, and then prompt for the required information: path and file name for sqlite; hostname, port, database schema and name, admin id and password for postgresql. 

You would be presented with an option on this screen to copy the existing database to the new. If you say "Yes", we will copy the data. If you say "no", you will get an empty database but with the full schema defined. If you say "yes" and there is any hint of an existing schema in the target db, we exit. If you want to just copy data, use the "Copy database" feature.

As part of the process for creating the new postgresql database, we would create a user and a password and store that. Since we are bound as a superuser during this process, we can toggle the flags as we need.

There would be a second new tool, something like "Copy database" which would allow you to specify a source and destination type. If the destination type is postgresql, then we will require the admin user and password. If the schema is not in the new database, then the process fails. If you want a new database, use the "new database" process. I think we also need to abort if there is any hint of data in there.

Once we have this tool, we change the options screen so all it does is swap between different databases. 

I think that will make things much cleaner, but I will have some work to do. I will work on this after we get the worst of the issues chased out of this PR.

matty0ung:
Ah, sorry, somehow missed that.

Seems there are two possibilities. One is that the data was inserted just fine and the "last primary key inserted" function is giving us "last automatically generated primary key", in which case we remove the assert and we're done.

Alternatively, if it's actually inserting a different primary key than we asked it to, then we need a bit more research to understand why.

M

Télécharger Outlook pour Android<https://aka.ms/AAb9ysg>
________________________________

mikfire:
I have a dirty thought. 

There's an odd bit of code in the original copyDatabase that talks about manually resetting the sequences in postgresql, (database.cpp, line 4717). As an example:
```sql
select setval(hop_id_seq,(SELECT MAX(id) from hop));
```
If I've understood, basically the sequence isn't invoked any time we actually supply the id. This is a known problem when doing mass inserts or copying databases. Well enough known that it is in the [PostgreSQL official documentation](https://wiki.postgresql.org/wiki/Fixing_Sequences).

If I've understood this correctly, your assertion will always fail. Until we manually update the sequence, the last inserted ID will always be 1. The random numbers I was seeing was that we are accessing a hash, so it always failed on the first insert but the actual ID was quasi-random. 

Of course, there's another problem here because I don't see we are currently ever updating the sequences. The first insert done after an import like that will fail. I think this should go in the schema helper part of the code, because this is the sort of meta-database thing SchemaHelper knows about.

Let me do some more testing to see if we can .. borrow .. the example given in the link above can just fix all the problems.

mikfire:
This one took me too long. There's an issue in ObjectStore::insertIntoJunctionTableDefinintion. We prepare the BtSqlQuery once, and try to reuse it a few times.

The problem is that BtSqlQuery::exec sets the bt_query to "" and the bt_boundValues to false as soon as it is used. This is fine as long as we only ever want to exec the query once. If we try to bind new values to the query, reallyPrepare() is called. Since bt_boundValues is now false, we call QSqlQuery prepare on bt_query. And Qt dumps core complaining that we are preparing an empty query

I think we should not be resetting the member variables.

mikfire:
My dirty thought was correct. You pretty much want to add this to DatabaseSchemaHelper::copyToNewDatabase(), immediately after the call to WriteAllObjectStoresToNewDb().
```c++
   // we need to rebuild all our sequences in postgresql
   if (newDatabase.dbType() == Database::PGSQL) {
      BtSqlQuery q{connectionNew};
      BtSqlQuery sequence{connectionNew};

      // so this query will return a list of queries we need to run. I have
      // taken this from https://wiki.postgresql.org/wiki/Fixing_Sequences and
      // have no idea how it actually works.
      QString resetIndexes = QString(
            "SELECT 'SELECT SETVAL(' ||"
            "       quote_literal(quote_ident(PGT.schemaname) || '.' || quote_ident(S.relname)) ||"
            "       ', COALESCE(MAX(' ||quote_ident(C.attname)|| '), 1) ) FROM ' ||"
            "       quote_ident(PGT.schemaname)|| '.'||quote_ident(T.relname)|| ';'"
            " FROM pg_class AS S,"
            "     pg_depend AS D,"
            "     pg_class AS T,"
            "     pg_attribute AS C,"
            "     pg_tables AS PGT"
            " WHERE S.relkind = 'S'"
            "    AND S.oid = D.objid"
            "    AND D.refobjid = T.oid"
            "    AND D.refobjid = C.attrelid"
            "    AND D.refobjsubid = C.attnum"
            "    AND T.relname = PGT.tablename"
            " ORDER BY S.relname;");
      if ( ! q.exec(resetIndexes) ) {
         qCritical() << Q_FUNC_INFO << "Could not find list of sequences to rebuild" << q.lastQuery() << q.lastError().text();
         return false;
      }
      while ( q.next() ) {
         QString doMe = q.value(0).toString();
         qInfo() << Q_FUNC_INFO << "doMe =" << doMe;
         if ( ! sequence.exec(doMe) ) {
            qCritical() << Q_FUNC_INFO << "Could not rebuild sequence" << q.lastQuery() << q.lastError().text();
            return false;
         }
      }
   }
```
or at least something like it. I got the first query from the previously linked Wiki page. It actually returns a list of queries to be run to update the sequences, which surprised me a little. I am willing to be I could be more clever, but this got it done.

The good news is that with promoting my db user to superuser, removing the assertion, modifying BtSqlQuery::exec to not reset the member variables and this change, I actually am able to import a SQLite db into Postgresql and add new things. Yay!

mikfire:
Unfortunately, something isn't right with respect to the inventory.
```
[13:34:56.684] (1dkgtd6phc) WARNING : QList<std::shared_ptr<QObject> > ObjectStore::getByIds(const QVector<int>&) const Unable to find object with ID  131  [database/ObjectStore.cpp:1157]
```
when I try to modify a newly added hop. 

It doesn't happen on the existing hops. 

The hop_in_inventory row was created, but the id of that row was never put back on the new hop. I haven't been able to understand how all of the insert work is done yet, so I am unable to provide a suggestion

matty0ung:
> There's an issue in ObjectStore::insertIntoJunctionTableDefinintion. We prepare the BtSqlQuery once, and try to reuse it a few times. ... I think we should not be resetting the member variables.

D'oh, of course.  Yes, you're absolutely right.  Will fix.

> I have a dirty thought [about PostgreSQL sequences].

This is the sort of thing that makes you want to have a database abstraction layer.  :grin: 

The "fix everything" SQL on the PostgreSQL looks pretty painful, but I wonder if we can do something slightly simpler.  If we could write a function along the lines of `void updatePrimaryKeySequenceIfNecessary(char const * const tablename)` then it could be a member function of Database with a no-op implementation for SQLite. WDYT?

> insertObjectInDb (need for superuser perms) - For now, I suggest we mark this as a known bug...

OK, that makes sense to me.  I like the idea of that QtKeychain library, so we might want to look at it in the future (for regular DB user passwords) but, for the superuser password, your approach of saying we should only ask for it when we need it, and not store it, feels like the right thing to do.

matty0ung:
Comments crossed in the post! :smile:
Will have a look at the inventory thing.

matty0ung:
> Unfortunately, something isn't right with respect to the inventory.

This is now fixed. (I was foolishly trying to get some Hop objects from the InventoryHop store.)

Various other fixes in the last commit.  Currently working on Brewtarget tooltip and the magic potion SQL to fix PostgreSQL sequences.

mikfire:
Two last problems that I can find, and I swear we can merge this.

In DatabaseSchemaHelper::migrate_to_8(), line 401. It would seem that when doing an insert from a select, you don't use the VALUES construct. It should look like:
```c++
migrationQueries.append({
      QString("INSERT INTO %1_in_inventory (%1_id) "
      // Everything has an inventory row now. This will find all the parent items that don't have an
      // inventory row.
      "SELECT id FROM %1 WHERE NOT EXISTS ( "
```
That took me longer to find that you might expect.

The second issue is in the same method, in the for loop starting at line 440. The yeast_in_inventory table does not have a column called "amount"; it has a column called "quanta". 

Fixing those two allows me to upgrade a database from brewtarget 2.3.1 to the current without an error. You can try this by getting the default_database.sqlite from the stable/2.3.1 branch, copying it into the right place and starting brewtarget.

I haven't tried this on postgresql, because I can't find an older backup I actually trust. I am not going to worry about that too hard right now.

matty0ung:
I am looking at a bug in ObjectStore that blows up when you import a new Style as part of importing a Recipe from XML.  I think it may be a transaction thing.  Watch this space.

mikfire:
There seems to be something weird with the mashs and mashsteps as well, but I am willing to handle it as a normal bug/pr. 

I am trying to create a new mash. I can create the mash, I can add a mashstep but then trying to use the mash wizard says "You need to add a step". If I restart brewtarget, that one step is gone. When I add the mash step, I see 
```
[09:53:56.405] (1dn7f0iebk) WARNING : QObject::connect: 
No such slot MashStepTableModel::changed(QMetaProperty,QVariant) 
 in /home/mik/brewtarget/matty0ung/src/MashStepTableModel.cpp:66  [:0]
```
My bet is that the view thinks it did a thing, but the signal wasn't properly caught so that database doesn't see the thing was done.

If I try to adjust an existing mash by deleting a mashstep, I get 
```
[09:51:50.930] (1dn7f0iebk) WARNING : 
  MashStep* Mash::removeMashStep(MashStep*) 
Tried to remove MashStep # 37  (from Mash # 34 ) but couldn't find it  [model/Mash.cpp:340]
```
I suspect this is because the step has actually been removed, but the table hasn't seen the signal to remove it.

matty0ung:
Right, I fixed my crash, then found another one and fixed that.  Have also made a change that should at least improve the issue you're seeing above.

Ideally, I'd like to go through the code and remove all uses of the SIGNAL and SLOT macros as they take away the compile-time checking you get when you use the address of the signal/slot function.

mikfire:
Mashes are still not working correctly. These errors do not seem to be in HEAD on develop, so I am thinking they are net new. I am also willing to merge as is, and handle this as a standard bug/pr.

The mash designer is not updating the "total collected wort" as I increase the infusion amount. If I say "Finish" in the mash designer, the steps are not shown in the mash table. If I just add a step, I see the step added in the mash table. Clicking "Mash Wizard" still says "There must be at least one mash step to run the wizard". If I restart brewtarget, there is nothing either in the table or the database.

I am very suspicious of the lack of an objectStoreWrapper::insert in MashStepEditor::saveAndClose. To make it work, I need to put the insert *after* the addMashStepToMash() which feels really, really weird. If I do it before, my database will suddenly have mashsteps where mash_id IS NULL.

I don't see that calling insert is setting cacheOnly to false, which I think it should be. I also see this not happening in the Equipment editor, but it seems to happen almost everywhere else. MashWizard is likely going to take a lot of work to figure out where inserts need to be called.

Trying the mash wizard after my fixes above generates an assert failure, saying:
```
(1do8u7eozk) ERROR : 
int ObjectStore::impl::insertObjectInDb(QSqlDatabase&, const QObject&, bool) 
Wrote new MashStep  to database (with primary key  51 ) but it already had primary key 49  
[database/ObjectStore.cpp:794]
(1do8u7eozk) ERROR : ASSERT: "false" in file 
/home/mik/brewtarget/matty0ung/src/database/ObjectStore.cpp, line 797  
[database/ObjectStore.cpp:797]
```
Trying that same thing in a slightly less messed up database does the same core dump, but with this message:
```
(1drdj14cg0) ERROR : 
int ObjectStore::impl::insertObjectInDb(QSqlDatabase&, const QObject&, bool) 
Wrote new MashStep  to database (with primary key  41 ) but it already had primary key 40  
[database/ObjectStore.cpp:794]
(1drdj14cg0) ERROR : ASSERT: "false" in file 
/home/mik/brewtarget/matty0ung/src/database/ObjectStore.cpp, line 797  [database/ObjectStore.cpp:797]
```

mikfire:
I tried to rework the signal/slot mechanisms a while ago. 

The problem was when we used overloaded slots. At that time, the only way to handle that was with the macros. They've since fixed that problem, or at least published a work around. 

I agree with fixing it, but that should be a different pull request, though. This one is gnarly enough as it is.

matty0ung:
Having a look at this.  Looks like, with the Mash Designer, we're creating a Mash OK but then inserting a MashStep in the DB without having set its Mash ID.

Similarly, I see what you mean about `MashStepEditor::saveAndClose` too.  It calls `Brewtarget::mainWindow()->addMashStepToMash()`, which calls `Mash::addMashStep()` but that doesn't save the MashStep in the DB.  Will have a think about where best to do that.  As you say, we don't want to try to insert things twice as it causes other problems. (We could bodge round them by calling `insertOrUpdate()` rather than `insert()`, but it would be nicer to just call insert once at the right time.)

matty0ung:
Whilst I'm looking at this, do you know what "Save Mash" (on the Mash tab of the current Recipe) is supposed to do? At the moment, it's saving a copy of the current Mash, but I wondered, if that's what it's for, whether it should be prompting the user to name the new Mash?

mikfire:
First, a disclaimer. This is a part of the code I've never really used, understood or particularly liked. I have long felt mashing in brewtarget is the hardest part to understand, and the part most in need of a significant redesign. Which I will do, just as soon as I get the brilliant insight as to what it should look like.

I believe the idea was that if you had a mash profile you always used -- like all my beers are usually single infusion, mashed at 66C for 60 minutes and initial temp around 20C -- you could create a named mash and just reuse that instead of having to build a new profile every time.

There is no separate window in which to create the mash profile. The save button was how we got around that problem. You could build the mash profile in the main window and then save it. You could also load that named mash, make changes as required and then save it and the next use of that profile would have the changes.

mikfire:
> Having a look at this. Looks like, with the Mash Designer, we're creating a Mash OK but then inserting a MashStep in the DB without having set its Mash ID.
> 
> Similarly, I see what you mean about `MashStepEditor::saveAndClose` too. It calls `Brewtarget::mainWindow()->addMashStepToMash()`, which calls `Mash::addMashStep()` but that doesn't save the MashStep in the DB. Will have a think about where best to do that. As you say, we don't want to try to insert things twice as it causes other problems. (We could bodge round them by calling `insertOrUpdate()` rather than `insert()`, but it would be nicer to just call insert once at the right time.)

I will point out almost every other editor window manually sets the cacheOnly flag after the insert call. EquipmentEditor doesn't, but I suspect that is not intentional.

What if we included a parameter to insert that would set the object's cacheOnly. I am thinking something like a boolean that defaults to false. I would expect the standard use case is that we want cacheOnly() set to false after the insert, so using that as the default would mean the least typing.

mikfire:
The issue in the MashWizard is that we are not calling either Brewtarget::mainWindow()->addMashStepToMash() or mashStep->setMash() when required. Assuming we want undo, we need the first call. If we don't care, then the second call. It's still a little broken with the sorting after adding new steps, but at least they show up and get sorted properly on restart.

The mash designer is supposed to clear the entire mash profile before it starts, which it doesn't seem to be doing. I will point out that the "Total Wort Collected" isn't working properly in HEAD on develop. So we will fix that as a separate problem.

matty0ung:
I must admit, I was a bit puzzled about the cacheOnly flag.  I see the effect but I wasn't sure about the purpose.  Originally I figured it was to mark an object as "this may not end up getting saved to the database" (eg we're creating a new thing and the user could either click save or cancel).  But then I thought we could infer that from the object's key being (or not being) valid, so I guessed there must be another use case for it, but I never quite followed through all the code paths to work it out.  If there's some simple rule about when it should and shouldn't be set, perhaps we could manage that somewhere centrally as you suggest.

mikfire:
I am dead against "inferring" anything. Inference like that will always break, as people forget or new people come on board. If we need a thing, we declare it and use it for that purpose. In general, I would rather things be easy to understand than to save even a few thousand lines of code. Code is cheap, my time is not.

Allow me to state that again. I _hate_ inferring status. If you need to know if a thing is this or that, use a specific and well-named flag to indicate if the thing is this or that. In this day and age, I do not really care if my in-memory foot print is now 1 byte bigger because I used a boolean where I could have inferred. Yes, there are problem domains (embedded systems, for example) where that really does matter. I am not coding brewtarget for those problem domains, and I am going to optimize my ability to read and understand the code over memory.

By the way. Have I mentioned how much I *hate* inferring status? It is almost always clever, and there are few things more dangerous in IT than people being "clever". It usually means you have made it difficult to impossible to ever change your approach, because there are all these side effects that simply are forgotten, unknown or otherwise not fixable. Writing code it intentionally rely on a side effect is just wrong.

The intent of that flag was mostly for speed. It provided a really fast way to create an object that did not require making one query to the database for each attribute for every object. As I recall, it required something like 10,000 - 20,000 queries just to load my database. It also had any number of weird side effects, because we would do things like recipe->recalcall, which would then rewrite attributes that we were going to read next. I could turn all that off by setting the cacheOnly flag. It literally dropped the number of queries by an order of magnitude.

It also became useful for the editors, so that we could create something in memory and not write it until the user clicks "save". Instead of writing each attribute as a separate statement, it allowed me to do one massive write that inserted everything. The fewer trips we make to the database, the faster performance. 

This had the unexpected but really pleasing side effect of not actually writing a thing to the database until the user said "Ok". It is difficult to describe how annoying it was to start adding a new hop, hit "cancel" and still have to delete the entry from the tree. This was honestly not planned, but was such a relief to me that it made the entire effort worth it.

The final reason is that there had been a previous attempt at some kind of caching mechanism. It was only ever applied to one of the base tables (fermentables, as I recall). When I was trying to make the networked database more performant, I had to deal with it. In studying the problem, I felt the first attempt was optimizing the wrong way. For the most part, we want one field flushed to the database immediately. Only in rare circumstances (like the editors) do we want to cache things and then do one massive flush at the end. This was my solution.

I believe I changed how I was doing things half way through and stopped using the set\*() methods on load. The idea still worked for those weird, calculated fields and it was so useful for the editors, so I kept it. I still think it is useful and am really set against changing it. No. Really. Do not break this part of the code unless you have a really solid replacement that does *not* infer status, because I really do not like inferring status.

matty0ung:
I think I get your drift. :smile: 

I'm pretty sure we violently agree that we should not add complexity to code in order to make some miniscule saving of memory footprint or run speed.  And I too don't like it when you have to jump through a lot of hoops to work out what state an object is in.

In this instance however, it seemed to me that, whatever the original history of the flag, there's nowa rather direct relationship between cacheOnly and "is stored in the database".  Keeping two separate ways of tracking the same thing risks creating complexity rather than simplicity.  Specifically, if we look at all possible of combinations of these two flags/states, we see the following:

- **New Object** state: `(key < 0 && cacheOnly == true)`. This is a new object that isn't yet stored in the database (key < 0) and for which we don't want setters to try to update the database (cacheOnly == true) as that wouldn't make sense.
- **Stored Object** state: `(key > 0 && cacheOnly == false)`. This is the state for most objects most of the time.  The object is stored in the database (key > 0) and we want calls to its setters to also update the corresponding data in the database (cacheOnly == false).  This means the in-memory object and its database record stay in sync.
- **Bad** state: `(key < 0 && cacheOnly == false)`. This state wouldn't make any sense. Having cacheOnly == false means we want setters to update the database, but they can't because key < 0 means there isn't a database record to update. If a setter were ever called on an object in this state I'd be expecting to log an error and fire an assert.
- **Dangerous** state: `(key > 0 && cacheOnly == true)`. This is where the object has a database record (key > 0) but we are asking the in-memory copy **not** to update the database when its setters are called (cacheOnly == true). This _might_ be OK if the user of the object is calling a lot of setters in succession and wants to optimise performance. As long as they always call `ObjectStoreWrapper::update()` at the end (and set cacheOnly to true again), then things will work. BUT, if that ever doesn't happen then the in-memory object will be out of sync with its database record and the user risks data loss. It seems like this state creates complexity that is only of benefit if we want some optimisation that is probably of only small benefit...

If we were to eliminate the cacheOnly flag, we'd reduce these four states down to two: **New Object** and **Stored Object**, which seems like a simplification to me.  In particular your points above are still addressed:

> we could create something in memory and not write it until the user clicks "save"

This would still be the case.  Calling setters on a newly-created object doesn't try to talk to the database.  Only once the user clicks "save" do we insert the object in the DB, give it a key, and thereby activate the behaviour where setters update the DB record.

> The intent of that flag was mostly for speed. It provided a really fast way to load an object that did not require making one query to the database for each attribute for every object. 

The ObjectStore code already does what we want here, without needing to reference cacheOnly.  When loading objects from the database, it does one query per table, reads an entire row at a time from from result set and uses that data to construct an object without calling its setters.  (This is the purpose of model/NamedParameterBundle.)  I don't know what performance is with PostgreSQL, but it certainly seemed fast enough with SQLite on my machine.


I'm _not_ saying we need to eliminate the cacheOnly flag as part of this commit, but, if we agree it's past its sell-by date then it would make fixing some of the other bits of the code simpler.

matty0ung:
OK, Mash and MashStep should be behaving a lot better now.  (I had missed finishing off the reworking of how the reference each other.)  Seems to work better on my local testing, but I'm not a guru of the Mash Wizard Designer things, so please shout if you see any oddities.

mikfire:
Sorry for the hiatus. There are issues with the mash steps and all the editors still.

The mash designer still does nothing. Saying "next" should add the step to the mashes, and it doesn't. It doesn't seem to add anything, but I haven't dug hard enough into the database to determine just what is or isn't happening.

Trying to run the mash wizard on a recipe that already has mashsteps dumps core like this:
```
#5  0x00005555557d9683 in NamedEntity::name (this=0x555556c0b8c0) at /home/mik/brewtarget/matty0ung/src/model/NamedEntity.cpp:225
#6  0x0000555555784f10 in MashStepTableModel::data (this=0x555556a52c30, index=..., role=0) at /home/mik/brewtarget/matty0ung/src/MashStepTableModel.cpp:244
#7  0x00007ffff73ae508 in ?? () from /usr/lib64/libQt5Widgets.so.5
#8  0x00007ffff73af465 in QItemDelegate::sizeHint(QStyleOptionViewItem const&, QModelIndex const&) const () from /usr/lib64/libQt5Widgets.so.5
#9  0x00007ffff73e2215 in ?? () from /usr/lib64/libQt5Widgets.so.5
#10 0x00007ffff73e242d in QTableView::sizeHintForColumn(int) const () from /usr/lib64/libQt5Widgets.so.5
#11 0x00007ffff739e25a in ?? () from /usr/lib64/libQt5Widgets.so.5
#12 0x0000555555784c76 in MashStepTableModel::mashStepChanged (this=0x555556a52c30, prop=..., val=...) at /home/mik/brewtarget/matty0ung/src/MashStepTableModel.cpp:203
#13 0x00005555555c76c2 in MashStepTableModel::qt_static_metacall (_o=0x555556a52c30, _c=QMetaObject::InvokeMetaMethod, _id=5, _a=0x7fffffffcbc0)
    at /home/mik/brewtarget/matty0ung/build/src/btobjlib_autogen/EWIEGA46WW/moc_MashStepTableModel.cpp:106
#14 0x00007ffff66002de in ?? () from /usr/lib64/libQt5Core.so.5
#15 0x00005555555d31c5 in NamedEntity::changed (this=0x555556cb53d0, _t1=..., _t2=...) at /home/mik/brewtarget/matty0ung/build/src/btobjlib_autogen/NLOY5YBGEN/moc_NamedEntity.cpp:249
#16 0x00005555557d9f39 in NamedEntity::propagatePropertyChange (this=0x555556cb53d0, propertyName=..., notify=true) at /home/mik/brewtarget/matty0ung/src/model/NamedEntity.cpp:353
#17 0x00005555557d41c2 in NamedEntity::setAndNotify<int> (this=0x555556cb53d0, propertyName=..., memberVariable=@0x555556cb5440: 1, newValue=1)
    at /home/mik/brewtarget/matty0ung/src/model/NamedEntity.h:373
#18 0x00005555557d32da in MashStep::setStepNumber (this=0x555556cb53d0, stepNumber=1) at /home/mik/brewtarget/matty0ung/src/model/MashStep.cpp:140
#19 0x00005555557cfba9 in Mash::impl::setCanonicalMashStepNumbers (this=0x555556906020) at /home/mik/brewtarget/matty0ung/src/model/Mash.cpp:57
#20 0x00005555557cb43a in Mash::removeMashStep (this=0x555556905ed0, mashStep=0x555556c0b8c0) at /home/mik/brewtarget/matty0ung/src/model/Mash.cpp:436
#21 0x000055555578d9b9 in MashWizard::wizardry (this=0x555556429460) at /home/mik/brewtarget/matty0ung/src/MashWizard.cpp:200
#22 0x000055555575d3d9 in QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (MashWizard::*)()>::call(void (MashWizard::*)(), MashWizard*, void**) (
    f=(void (MashWizard::*)(MashWizard * const)) 0x55555578d66c <MashWizard::wizardry()>, o=0x555556429460, arg=0x7fffffffd0d0) at /usr/include/qt5/QtCore/qobjectdefs_impl.h:152
```
I've edited that a little, because the first 4 entries really weren't interesting.

I am testing with a simple infusion and two batch sparges. If there are no sparge steps, two steps get created and the math looks roughly correct. Rerunning the mash wizard at that point dumps core. If I restart brewtarget there is one batch sparge step and the infusion step. Running the mash wizard dumps core again, in the same spot. Running brewtarget again shows just the infusion and no sparges. Since there are no sparge steps, two steps get created. And so on.

Attempting to remove a mash step by clicking the "-" button causes the same core dump, so I am thinking removing a mash step is causing something grief. It doesn't seem to happen with any of the other tables. Based on the core dump, I suspect you may be trying to get a name that doesn't exist any longer?

It would also be nice to fix the sort. The infusion step should be before the sparge steps, but I don't know if that is a related issue.

matty0ung:
Thanks.  This is all good stuff.  The fixes are all simple but one needs to touch a fair bit of code, so I'll work on it over the next couple of days.

The one that involves lots of small changes is to do with object ownership.  Essentially when we really really delete something - eg a MashStep that has no meaningful existence once it has been removed from a Mash - we should really be using shared pointers properly to make sure we can the remove the object from the DB and ObjectStore but still have it in the undo/redo stack in case the user wants to undo the deletion.  The shared pointers are all already there in the ObjectStore, I just need to get a few more bits of the code to use them.  (Mostly where we come back out of the database layer into the existing code, we just reach inside the shared pointer to use the raw pointer, because everything was originally done with raw pointers (and which is still correct when you're creating a Qt object that the framework owns by some chain of parenthood etc).  Anyway, this use of raw pointers works fine when we're not deleting things or are only soft deleting things or can recreate the things that are deleted, which is ~99% of the code.  But I need to make more things pass the shared pointer around to ensure the right things happen in the other ~1% of cases.  A side benefit of this should be that we clear up a few potential memory leaks in the existing code.)

matty0ung:
OK, things should be somewhat more robust now.  I'm still uncertain about some of the finer points of the Mash Wizards and Designers (eg exactly when and how it should complain about your mash being too thick for it to be able to proceed) but I don't _think_ this is related to the database layer!

mikfire:
The mash designer still isn't behaving properly.  I think this should be an easy fix and I am willing to sign off on this merge as it exists, as long as we put a problem ticket in so we don't forget to fix this.

On starting the mash designer, it is supposed to delete all the existing mashsteps. Currently, it doesn't. The behavior is very strange after that, as it seems the designer is modifying the existing steps instead of creating new ones?

If I manually remove the existing steps, it behaves more as I would expect. The total collected wort is wrong. That will be a signal, I am quite certain. I've broken this screen enough times I could almost tell you what signal it is.

mikfire (approved):

Issue #613, opened 2021-07-12
More options on ABV calculation


matty0ung:
Alcohol Tool currently only offers one simple method of calculating ABV.  There are more sophisticated (= presumably more accurate) algorithms, eg the official one used in the UK as described here: https://www.gov.uk/government/publications/excise-notice-226-beer-duty/excise-notice-226-beer-duty--2#calculation-strength

mattiasmaahl:
Cool, do you think we should implement it in a way that lets the user decide what method to use? or do we implement the one that we figure is the best one?

matty0ung:
Personally, I'd want to use whatever we think is the most accurate, rather than ask the user to choose from some list of algorithms that they likely know nothing about.  (AIUI both the methods here https://www.brewersfriend.com/abv-calculator/ overstate alcohol content for beers with low-to-medium starting gravities.)  I'm happy to be schooled in what "most accurate" means(!), but my starting assumption was the UK tax authorities would have some stake in accuracy because it affects how much alcohol duty they collect from professional brewers!

mattiasmaahl:
You make a good point, then we just have to decide which is more accurate! Sorry, cannot help in that regard, but can do some research into the matter and inform of my findings.
Lets discuss this futher. 😊

mattiasmaahl:
After doing some research I found a bunch of "homebrew" variants of calculating the ABV, some actually kinda close to what you link to above.  
I also found a paper detailing a 'few' calculation formulas and discussing the differenses and accuracy of them. [https://onlinelibrary.wiley.com/doi/pdf/10.1002/j.2050-0416.2009.tb00387.x](https://onlinelibrary.wiley.com/doi/pdf/10.1002/j.2050-0416.2009.tb00387.x)  
That paper mentions your British legislated values as well, but it goes in to how to calculate the % Alcohol by weight (A<sub>w/w</sub>) and discusses that in great detail, that on the other hand is too complicated I think. although they do provide formulas for converting between SG and RE they do not discuss converting between OG and OE.

I think we should look into the above British formula constants and see if we can find a way to automatically select the correct constant for calculating the ABV.

matty0ung:
Good research!
I think there's a relatively neat way to pull out the right constant for the UK formula.

mattiasmaahl:
[...]

PR #435, opened 2018-10-15
Removed Default Amount field from several editors 

jhmcstanton:
Attached are screenshots of the change + local build
fermentable_editor_update
hop_editor_update

misc_editor_update
yeast_editor_update
local_build

PR #631, opened 2022-01-27
Refactor Unit, UnitSystem etc code - groundwork for BeerJSON

matty0ung:
As mentioned in https://github.com/Brewtarget/brewtarget/discussions/621, this is a chunk of work to make Unit, UnitSystem and so on more consistent and extendable, which will help a lot in implementing support for BeerJSON (which requires us to support more different units in more different circumstances than we do right now).

I did the work first in Brewken (where you can see a bit more of the JSON groundwork) and in merging it across to Brewtarget, I have also brought along a fix for https://github.com/Brewtarget/brewtarget/issues/613 and a proper merge of https://github.com/Brewtarget/brewtarget/pull/435. I have also brought across the enhanced version of the Alcohol Tool (which allows you to do temperature correction and alcohol calculation in one step, saving time and removing rounding errors from the previous approach).

In doing the refactoring, I have made the pop-up menus for selecting per-field systems of measurement a bit more visible (you get a mouseover indication in a lot more cases and can now often left-click as well as right-click to get the menu).  I have also fixed a bug where the scale sub pop-up menu would not be updated after you've changed a field's system of measurement.

I have removed one feature, which seemed unlikely to be used, but which could, if needed, be added in a future fix/enhancement.  Although you can still select date format globally (from the Options dialog), you can no longer set the date format on a per-field basis.  (This is because date format is not really a system of measurement!)  It seemed unlikely to me that someone would want different date formats in different places, but if I am wrong about this, it would be small-ish work to reimplement per-field date format selection.

I have tried to do a fair bit of testing, and have added some initial unit tests for converting between measurements, but, as ever, please shout if you spot anything I have missed.

Although a lot of files are touched, the main changes are in the measurement directory & Measurement namespace.   Unit and UnitSystem remain somewhat as before, but SystemOfMeasurement is now pulled out from Unit to be freestanding, and PhysicalQuantity is created.  See comment in `measurement/PhysicalQuantity.h` for explanation of the structure.  Amount is a new small wrapper class that makes it easy to pass quantity + units around (eg {5.0, Measurement::Units::pounds} is 5.0lb).  This was helpful to me in debugging and will I think be useful in the future. 

I've also pulled out a lot of measurement/formatting/localization/config stuff from brewtarget.h/brewtarget.cpp to put them in more appropriate places.


mattiasmaahl (approved):
Looks good to me, It builds and runs on windows, all actions are green. I'll Merge this into develop.

PR #633, opened 2022-02-02
Small typo in CmakeLists.txt

mattiasmaahl:
In Cmake Lists there was an include to Brewken.h instead of brewtarget.h
 
It's not a showstopper, but to keep things in order. :)


matty0ung (approved):

mattiasmaahl:
Haha, well it was, as we in the business would say, an easy fix! 🤣😇

Issue #634, opened 2022-02-08
Crash when creating new Equipment


mattiasmaahl:
So I was creating a new Equipment.

To replicate:
1. go to Equipment tab
2. right-click on the EquipmentFolder
3. select new equipment from pop-up menu
4. fill in the various details, I changed the Physics->"Mash tun Volume" and <EQUIPMENT NAME> -> Batch Size to match the boiler in mind.
5. hit save, then it crashes.

Version, Brewtarget:develop (my own local build of the latest develop)

Brewtarget-log of the crash:
[brewtarget.log](https://github.com/Brewtarget/brewtarget/files/8023877/brewtarget.log)

Here's the code that Crashes: `src/UiAmountWithUnits.cpp:158~167`

~~~
Measurement::Amount UiAmountWithUnits::toSI() {
   // It's a coding error to call this function if we are not dealing with a physical quantity
   Q_ASSERT(std::holds_alternative<Measurement::PhysicalQuantity>(this->getFieldType()));  <-- this line!
   Q_ASSERT(this->units);

   return Measurement::qStringToSI(this->getWidgetText(),
                                   this->units->getPhysicalQuantity(),
                                   this->getForcedSystemOfMeasurement(),
                                   this->getForcedRelativeScale());
}
~~~

@matty0ung you have any idea what this could be?

matty0ung:
I'll have a look either tonight or tomorrow.  If you have the stack trace that will make the root cause obvious (I hope).  If not, don't worry, as I'm sure I'll be able to reproduce on Linux.  It's probably a relatively straightforward fix.  (Famous last words.)

mattiasmaahl:
Sorry, No stacktrace available. 😞

the rows in the logfile above are the last rows in the logfile before the crash. all rows before those are from startup of the app.

matty0ung:
Had a quick look.  I think this is is because of `lineEdit_hopUtilization->toSI()` in `Equipment::save()`.  It's not meaningful because the amount is a percentage, not a measure of a physical quantity, and therefore there is no SI unit for it - hence the assert.  The fix is that, for this field, we should just use it without calling `toSI()`.  There may be one or two other fields with similar problems.  Will check and do a patch tomorrow hopefully.

matty0ung:
The short fix is in https://github.com/Brewtarget/brewtarget/pull/635.  Will have a think about something that would flush these issues out at compile-time.

mattiasmaahl:
[...]

PR #635, opened 2022-02-09
Fix crash when creating new Equipment

matty0ung:
Fix for https://github.com/Brewtarget/brewtarget/issues/634 plus more general tweak so we get an error message rather than an assert when trying to call `toSI()` for a BtLineEdit that does not related to a physical quantity.  (Though we should fix the inheritance hierarchy for BtGenericEdit etc so that these things get flushed out at compile time.)


mattiasmaahl (approved):
This works as expected, I will merge this and close the issue.

Issue #620, opened 2021-10-14
Mash Designer misbehaving


matty0ung:
New DB layer has introduced a bug that means the Mash Designer does not work as designed.  There are two issues:

1. On starting the mash designer, it is supposed to delete all the existing mash steps. Currently, it doesn't. The behaviour is very strange after that, as it seems the designer is modifying the existing steps instead of creating new ones? If you manually remove the existing steps, it behaves more as we would expect.
1. The total collected wort is wrong. @mikfire says "That will be a signal, I am quite certain. I've broken this screen enough times I could almost tell you what signal it is."

I will work on a fix for this once https://github.com/Brewtarget/brewtarget/pull/617 is merged.

matty0ung:
I have a fix for •1•.  Looking at •2•.

matty0ung:
I think this is now fixed, so closing.

PR #638, opened 2022-02-19
Fix for Mash Designer problems (https://github.com/Brewtarget/brewtarget/issues/620)

matty0ung:
There were a few places where we were still trying to do things directly with `MashStep` objects rather than just giving them to the owning Mash and letting all the logic be centralised in `Mash::addMashStep()`.  Similarly, part of the original bug was trying to grab some info from a newly-created `MashStep` before it had been added to the Mash.  We now just wait for the Mash to tell us about changes to its MashSteps, per comment in `src/tableModels/MashStepTableModel.cpp`

Switched more of the code to use shared pointers which removes some memory leaks and makes undelete work even on things that have been "hard" deleted (because the shared pointer keeps the object in existence even after it was removed from the `ObjectStore`).

Tried to remove some of the duplicated code from the files in the tableModels directory, though there are some limitations about what we can do here with the Qt MOC not being able to handle C++ templates (see comment in `tableModels/BtTableModel.h`

### DIFF
--- a/src/BrewDayFormatter.cpp
+++ b/src/BrewDayFormatter.cpp
@@ -235,7 +235,7 @@ QString BrewDayFormatter::buildInstructionHtml() {
              .arg(tr("Step"));
 
    QList<Instruction *> instructions = recObs->instructions();
-   QList<MashStep *> mashSteps = recObs->mash()->mashSteps();
+   auto mashSteps = recObs->mash()->mashSteps();
    int size = instructions.size();
    for (int i = 0; i < size; ++i) {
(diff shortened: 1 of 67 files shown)


matty0ung:
Need an extra include or two for Mac build - will add this weekend!

matty0ung:
Since it's been a few weeks there are folks waiting on this fix, I'm going to merge it.  If anyone does want to do a code review then, obviously, that would be great to have, even subsequent to the merge, and I can incorporate feedback into an upcoming patch.

Issue #641, opened 2022-03-22
Fixed efficiency


RSP1000:
Hello there  :)
I used to be able to change the value of the expected efficiency in the window composing a recipe. Now the value is fixed at 70% and cannot be changed. I'm using the latest version from git on linux. Does anyone else have a similar problem?.

Have a good one.

matty0ung:
OK, I think I have a fix. `BtLineEdit::onLineChanged()` needs to `emit textModified()` in the event that the field does not hold a physical quantity.  It's a small change.  I'll do a patch either tonight or tomorrow morning.

matty0ung:
Patch uploaded at https://github.com/Brewtarget/brewtarget/pull/642.  Awaiting review.

PR #642, opened 2022-03-23
Fix for https://github.com/Brewtarget/brewtarget/issues/641

matty0ung:
This fixes the issue of recipe efficiency not being modifiable.  (It's a bug from the changes I made in https://github.com/Brewtarget/brewtarget/pull/631 where a signal didn't get emitted.)

It should be possible to merge this either before or after https://github.com/Brewtarget/brewtarget/pull/640 and https://github.com/Brewtarget/brewtarget/pull/638.


mattiasmaahl (approved):
Looks good

Issue #645, opened 2022-04-02
A zombie recipe ;), just cosmetic thing.


RSP1000:
Please create a recipe, add just one kind of malt (for testing purpose) then delete the recipe and do nothing else, just exit BT.
Once again start BT, on the right side of the screen you'll see the content of deleted recipe in all its glory, 
but it should be gone for good now, right?.The recipe was deleted and program restarted, why it is still there.
Having déjà vu, a glitch in matrix :), I thought I had successfully removed it. Just a cosmetic thing.
Regards :)

matty0ung:
Ah, good spot!  Yes, in many circumstances, when items are deleted in Brewtarget they are just tagged with a "deleted" flag rather than actually removed from the DB.

What should happen is that, when you delete the currently selected recipe, we should select another one - perhaps the first in the list. Otherwise the deleted recipe remains displayed and editable - including after closing and restarting the program, as you've spotted.

Hopefully this is a small fix.  I'll have a look...

matty0ung:
From a bit of looking, I _think_ the issue only arises when the recipe you're deleting is the last one in the list.  If you delete a recipe that's not last in the list, then the next one in the list becomes the selected one straight away.  So, probably just need to add some logic to try to select the previous one when the last one is deleted.

RSP1000:
Hello
Good Sir from my perspective, some more logic sounds good :).
What about idea that recipe composer window will get open with default values
( zeroed, no ingridients etc.) in case when there are no available recipes. Well, it’s just a thought. 
Anyways, from now on it can be only better :).
As always thanks for explanations and good job.

matty0ung:
Just to let you know, I found the issue.  There is already logic in `MainWindow::deleteSelected()` to select a new recipe after one is deleted, and to return to the top of the list when we deleted the last recipe.  However, it needed a small enhancement to deal with the case where there are folders.  When you go back to the top of the "tree" of recipes, if the first item is a folder, you need to keep skipping through the list until you find an actual recipe or you get to the end of the tree.

I've coded a quick fix that will make its way into the code at some point and improves things in many cases.  Essentially, when you delete the end recipe, it looks for the first recipe that's not in a folder and sets that as the selected one.  However, there is another edge case that the fix doesn't cover, which is where all your recipes are in folders.  This needs a slightly bigger fix in a different part of the code - to look through the full list of recipes including ones in folders.  I've made a note in the code to come back to that at some point.

[...]

PR #655, opened 2022-08-14
Build fixes etc for Release 3.0.0 (was 2.4.0)

matty0ung:
I've changed CMakeList.txt, src/CMakeList.txt and QtDesignerPlugins/CMakeLists.txt so that all the build logic and settings are in the first of these 3 files, and the latter 2 just contain lists of source files.

I've turned on more of the CMAKE_AUTOxxx settings and removed all the (now unnecessary) places where we were manually listing .ui files and .h files that need MOC processing etc.

Still in CMakeList.txt, I've tried to put things together that go together and improve commenting.  For Linux, I've also got package building (`make package`) and package lint (`make package_lint`) seemingly working, but it could do with some proper testing.  Will probably need some help getting Windows and Mac packaging working, as I'm currently just pushing commits up to github to see what happens when I change those bits of CMake/CPack, which isn't very productive... :smile: 

On Linux the build process now takes `CHANGES.markdown` and converts it to a Debian-format changelog that can be included in the .deb package.  (The lintian program is quite strict about formats, so there's a bit of not-very-pretty scripting here.  It also doesn't like spelling mistakes, so I've corrected a couple of typos in issue descriptions!)

`CHANGES.markdown` mentions all the closed issues that were modified since the last release in 2016.  Could probably do with going through that and filtering out ones that were closed without code change.

Have also:

- Merged a fix for https://github.com/Brewtarget/brewtarget/issues/645 across from Brewken (because the issue got closed so the fix ought to be in the next release).
- Made a fix for https://github.com/Brewtarget/brewtarget/issues/656 (as it was a relatively straightforward fix for something straight not working).
- Removed `/src/BrewDayWidget.cpp`, `/src/BrewDayWidget.h` and `/ui/BrewDayWidget.ui` as they do not seem to be referenced anywhere and were presumably replaced by BrewDayScrollWidget.xxx a while ago.
- After commenting out `build-scripts/linux/apply-patches.sh` which was breaking the build on Ubuntu 18.04, I then removed the entire `build-scripts` directory (and the sole reference to it) as the contents are no longer used or needed AFAICT.


mattiasmaahl:
Will try to get windows build tested during the week. Will revert.

matty0ung:
Still fighting with the Windows build.  I've managed to get the build process itself not to barf out obscure errors (by which I mean errors that nobody else seems ever to have reported having!), but I haven't yet worked out how to get `make install` to put all the DLLs in the right place.

matty0ung:
I _think_ this is now working.  If anyone has time to check the Windows and/or Mac packages, that would be great.  If not, we could at least push the Linux release live and then do any further fixes in 3.0.1.

matty0ung:
Eliminated the need to run `make changelog` as a separate step.  Builds are now producing some sort of package on all 3 platforms. Merging everything so we can do the release.

Issue #670, opened 2022-10-19
Turkish Language selection in settings not working


smntsn:
First of all I'm using this great software about 2 years and I LOVE IT :) THANK YOU :)
My problem is (also for 2x versions) Turkish language selection changes nothing (Interface still in English).
I fixed it by downloading Turkish translation>converting to .qm with QtLinguist and replacing it in installation folder(for 2x version).

But, now I can't do it. 

I don't know if supplied "bt_tr.qm" is not Turkish or setting page language selection is not working.

Thanks in advance. 

![Ekran görüntüsü 2022-10-19 174931](https://user-images.githubusercontent.com/13038415/196725333-395bd870-2c59-400a-91d6-79696f47efef.jpg)

matty0ung:
OK, so I've fixed the build so that it is properly synchronising the "translations/*.ts" files (ie translation source files) with the translatable strings in the source code.  Once this fix is merged, you should be able to make useful edits to translations/bt_tr.ts.

I don't know whether generating the .qm file from QtLinguist and manually installing it is always guaranteed to work, but either way if you share your translations with us then we'd be very happy to incorporate them in the main build.

matty0ung:
For now, I've put a "placeholder" translation in for "About Brewtarget" so it says "Turkish for about Brewtarget".  This means that, by showing the "About" dialog, we can convince ourselves that "display in Turkish" is picking up strings from the right place.

Once we have some Turkish translations in https://github.com/Brewtarget/brewtarget/blob/develop/translations/bt_tr.ts then we won't need that placeholder.

matty0ung:
Sorry, jumped the gun a bit.  Wait for https://github.com/Brewtarget/brewtarget/pull/672 to be merged.

matty0ung:
We'll do a 3.0.2 release soon, which will contain the fixes mentioned above.

matty0ung:
The [3.0.2](https://github.com/Brewtarget/brewtarget/releases/tag/v3.0.2) release is now out, so I'm closing this ticket.  But please don't hesitate to re-open it (or raise another issue) if you hit problems with that new release.

PR #671, opened 2022-10-21
Fix build system to properly update translation source files

matty0ung:
This is a step towards fixing https://github.com/Brewtarget/brewtarget/issues/670

I'm doing two things here:
 - Discarding the ".po" translation source files and deciding that the ".ts" files are the master translation files.  (The ".ts" files are the Qt native format.  We only needed the ".po" ones when we were using Launchpad.
 - Updating the build system so that it automatically adds new entries to the ".ts" files when new translatable strings get added to the .cpp or .ui files.  (Essentially this is equivalent to running `lupdate` on all the ".ts" files when we do make

Of course this is all the easy bit and actually getting the translations done is a whole other issue, but it's a step in the right direction I hope!


mattiasmaahl (approved):

matty0ung:
There's another fix to come here BTW.  Just realised we're only getting the strings from cpp files and not the ui ones.  But at least I know why! :smile:

PR #686, opened 2022-11-27
Fixed release notes

matty0ung:
I'll get this working in a minute!


matty0ung:
See https://github.com/Brewtarget/brewtarget/pull/687 for what I was trying to do!

PR #692, opened 2022-12-20
More tweaks to Mac build

matty0ung:
This is a first attempt to get the Mac build to do static linking -- seeing as how we have had such problems with getting all the right dynamic libraries in all the right places when creating bundles with CPack.


matty0ung:
I'm going to merge this on the grounds that it can't make things worse for the Mac package.

PR #698, opened 2022-12-29
Putting the title back in brewtarget

### LINKED ISSUE #697: Brewtarget logo is missing from main window
Normally, below the big blue plus and minus symbols there should be the red "target" Brewtarget logo and the word "Brewtarget" in two shades of brown (eg see the screenshot in https://github.com/Brewtarget/brewtarget/issues/609).  I've just noticed on my local build that it's not there any more.  I'm guessing it's some one-line error somewhere (and quite likely as an oversight on my part), but, off the bat, I can't see where.  Eg I had a look at mainWindow.ui and brewtarget.qrc and couldn't see what might have changed (or when) for this to disappear.

mikfire:
Closes #697. 

The place to looks for the images is in brewtarget.qrc. It looks like the file got reworked a while back, and the two title images were dropped.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,8 @@ build/*
 
 .vagrant
 *.user
+
+# ccls files
+.ccls-cache/
+src/.ccls-cache/
+src/compile_commands.json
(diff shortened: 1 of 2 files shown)


matty0ung (approved):

matty0ung:
Thanks @mikfire ! :+1:

PR #706, opened 2022-12-31
Water chemistry misbehaving -- closes #705

### LINKED ISSUE #705: Water chemistry was misbehaving
I use the water chemistry feature heavily, and it wasn't working.

It came down to two major issues. The first was, as we discussed in another thread, the intializer for the Salt was not right. amount_is_weight, amount and addTo just are miswired.

The other problem was that BtTableModel::removeDuplicate was removing both duplicates and things marked as either deleted or to not be displayed. I am uncertain of that logic, but it certainly wasn't right for Salts as they are always marked as not displayed.

A pull request will be made directly.

mikfire:
I had to fix the initializer for the Salt object. The amount_is_weight, the amount and the addTo fields were all kind of cross wired.

The harder problem was the removeDuplicates method. I was unsure of the side effects of removing the deleted and not displayed tests, so I created a new method called `removeDuplicatesIgnoreDisplay` and used that in the SaltTableModel.

Things seem to be working better now.

### DIFF
--- a/src/BtTreeModel.cpp
+++ b/src/BtTreeModel.cpp
@@ -1411,7 +1411,7 @@ void BtTreeModel::observeElement(NamedEntity * d) {
       connect(d, SIGNAL(brewDateChanged(QDateTime)), this, SLOT(elementChanged()));
    } else {
       connect(d, SIGNAL(changedName(QString)), this, SLOT(elementChanged()));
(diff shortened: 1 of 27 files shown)


matty0ung (changes requested):
All looks good.  On line 60 we also want to change `type` to `isAcid`: 
```
 m_is_acid         {namedParameterBundle(PropertyNames::Salt::isAcid        ).toBool()} {
```

matty0ung (approved):

Issue #717, opened 2023-02-04
Hop editor crashes App


Melonbob:
The installed default database from developer branch builds doesn't coincide with the DefaultData.xml at least with hop records. The db is leaf and the xml is pellets. How do I rebuild it from the xml?  

Also when I went to edit with the Hop editor window, the pull-down menus were duplicated. The top set crashes the App while the bottom set does not.

![image1](https://user-images.githubusercontent.com/52466784/216731891-7a308136-a8e3-4428-b78a-406b53ee64ad.jpeg)

![image2](https://user-images.githubusercontent.com/52466784/216732609-5ec9a16c-3691-4e0d-8cb6-79e7ef842b76.jpeg)

![image3](https://user-images.githubusercontent.com/52466784/216732676-a2a85936-de57-42e2-805c-5c400d7183ec.jpeg)

matty0ung:
Ooh, I thought I'd fixed this, sorry.  It's a bug caused by when I accidentally brought to much of the "in development" BeerJSON work into the dev branch.  I'm travelling at the moment but will try to get a fix in the coming week.

matty0ung:
So I've done a fix for the UI bit.  The data import bit probably needs a bit more looking at.  Do you see any errors in the logs when you start up with a clean database?  What is supposed to happen is that the code should read the contents of `DefaultData.xml` into the DB using the regular XML import process.  If that fails for some reason then, usually, there should be some error message in the logs.

(The reason all this is being changed for BeerJSON is that it recognises additional forms of hops (Extract, Wet Leaf and Powder) that don't exist in BeerXML and therefore we didn't previously support.)

Melonbob:
I put the log here from startup here.  Looks like a clean, normal startup. 
https://github.com/Brewtarget/brewtarget/issues/714#issuecomment-1418141222

PR #718, opened 2023-02-04
Remove duplicate combo box entries on Hop editor

matty0ung:
Fix for https://github.com/Brewtarget/brewtarget/issues/717


matty0ung:
I'm going to merge this as I think it's not too controversial.

PR #722, opened 2023-02-26
Add Meson/bt as alternative build and packaging system; make enum deserialisation more robust

matty0ung:
The enum deserialisation is just adding a fallback for case-insensitive matching.

The Meson build stuff is mostly in two new files at the top-level directory: `bt` and `meson.build`.  The former is a Python script that can:

- install almost all the necessary dependencies for building the software
- set up the Meson build environment (in the `mbuild` subdirectory)
- after a Meson build is done (via `cd mbuild; meson compile`), package up the software (into `mbuild/packages` subtree)

Note that the current CMake/CPack build is pretty much unaffected by this new work.  You can still run all the regular CMake stuff in the `build` subdirectory.  CMake/CPack builds should continue to function as correctly as they do now (ie just fine on Windows and Linux, and fine except for package building on Mac).


penguinpee:
Nice! I'll give that a try as the build backend for the Fedora package. Are there plans for switching to Meson completely?

matty0ung:
> Nice! I'll give that a try as the build backend for the Fedora package. Are there plans for switching to Meson completely?

There are still a few glitches to iron out, but once things seem like they are working, we'll see what folks think.

Personally I think using Meson simplifies the build, partly because Meson doesn't try to do all the things that CMake does, so you end up doing all the "not-build" things in a regular scripting language.  And, for non-trivial scripting, I think Python is nicer than bash.  Of course moving away from CPack it makes packaging less automagical, but CPack has been far from trouble-free, and the upside of doing things more explicitly is that we then have more control over library dependency and other packaging joys.  Certainly doing the scripting has helped me understand a lot better how packaging works on Windows and Mac!

meson.build
```diff
@@ -0,0 +1,1449 @@
...
+compiler = meson.get_compiler('cpp')
+message('Using Meson', meson.version(), ', ', compiler.get_id(), 'compiler and', compiler.get_linker_id(), 'linker')
+
+#
+# We need two versions of the main executable name because they are different on Windows:
+#    - We need a "target" name _without_ the '.exe' suffix for the executable() command because, on Windows, Meson will
+#      always add its own '.exe' suffix and we don't want to end up with '.exe.exe' as the suffix!
+#    - Everywhere else, we want the actual name of the file, including the '.exe' suffix.
+#
+# The default (on Windows and Linux) is to use Unix-style executable names, ie all lower case
```
eli-schwartz, line 210: Do you need the basename or is the full filename okay? You can calculate the target name, then take the returned `executable()` object and use its `.full_path()` method to acquire the string value of the output filename, prefixed by the full builddir path.

meson.build
```diff
@@ -0,0 +1,1449 @@
...
+
+#=======================================================================================================================
+#============================================== Frameworks and Libraries ===============================================
+#=======================================================================================================================
+#
+# It would be neat within the build system to automate the _installation_ and upgrade of libraries and frameworks on
+# which we depend.  However, I have yet to find a universal pain-free solution.
+#
+# Meson has its own dependency management system, Wrap, but the list of pre-provided projects at
+# https://mesonbuild.com/Wrapdb-projects.html does not include Qt, Boost, Xerces, Xalan or Valijson.
```
eli-schwartz, line 246: Qt at least would likely be far, far too painful to reimplement.

Anything else is probably feasible, and you can submit a request or a PR to add it to the WrapDB. We'll be happy to accept it. :)

meson.build
```diff
@@ -0,0 +1,1449 @@
...
+   libcVersion = lddOutputLine1.split(' ')[-1]
+   message('libc version is:', libcVersion)
+   if libcVersion.version_compare('<2.34')
+      needLibdl = true
+   endif
+endif
+dlDependency = dependency('dl', required : needLibdl, static : true)
+#
+# On newer versions of Meson (0.62.0 and newer), the dependency has special-case code for dealing with libdl being built
+# into the compiler.  This means on Mac, for instance, we can have dlDependency.found() is true without there being any
```
eli-schwartz, line 420: This is actually in large part why Meson has a dl dependency. You shouldn't need to mess with ldd -- because `dependency('dl')` will check for the dl functionality with two possible "dependency providers":
- check to see if `dlopen()` is available in libc
- check to see if an external `dl` library is available

But on versions of Meson before 0.62.0, it will simply not find anything at all, even if needLibdl is true. So Meson will just fail in that case.

Granted, if a case-sensitive cmake `Finddl.cmake` exists then meson will fall back on that for meson versions <0.62.0, but I am not aware of cmake itself shipping with such a thing.

If you need to support meson 0.60.0, then I'd suggest not bothering with ldd at all, and just using this:

```meson
if cc.has_function('dlopen', prefix: '#include <dlfcn.h>')
  dl_dep = declare_dependency()
else
  dl_dep = cc.find_library('dl', has_headers: 'dlfcn.h')
endif
```

meson.build
```diff
@@ -0,0 +1,1449 @@
...
+                                            dirs : [libbacktraceDir])
+
+#======================================== Find the other libraries we depend on ========================================
+#
+# See https://mesonbuild.com/Reference-manual_returned_dep.html for what info we can pull from a dependency object
+#
+# For BeerXML processing we need Xerces-C++ and Xalan-C++.  Meson can find both of these automatically using CMake's
+# find_package(), as long as (a) CMake is installed(!) and (b) we provide the right library names ('XercesC' per
+# https://cmake.org/cmake/help/latest/module/FindXercesC.html and 'XalanC' per
+# https://cmake.org/cmake/help/latest/module/FindXalanC.html)
```
eli-schwartz, line 448: Probably the official way to search for both is with `dependency('xerces-c')` and `dependency('xalan-c')`. xalan-c provides a cmake package, but xerces-c doesn't.

Both provide pkg-config files which are lowercased and have a hyphen, so you need to use the meson 0.60.0 support for `dependency('xerces-c', 'XercesC')` with multiple possible names for the same dependency call, if you want to be able to find either one. If you can find it with pkg-config, you don't need cmake installed. You can get the 'libdir' variable from the dependency if its `.type_name() == 'pkgconfig'`.

meson.build
```diff
@@ -0,0 +1,1449 @@
...
+               output : 'manpage.1.md',
+               configuration : exportedVariables,
+               install : false)
+
+#
+# We also want to inject build-system variables into various platform-specific packaging scripts, control files and so
+# on.
+#
+# We ultimately want the generated files inside the 'packaging' subtree of the the build directory, but that is not
+# permitted by configure_file, so, that gets done by the bt script.
```
eli-schwartz, line 1221: You can, if you use `subdir('packaging')` and then define the configure_file() inside packaging/meson.build.

meson.build
```diff
@@ -0,0 +1,1449 @@
...
+   # Note that, although man pages are frequently compressed, the exact compression mechanism is distro-specific, so
+   # Meson now considers such compression outside the scope of the build system.  (We therefore do it in the bt build
+   # tool script.)
+   #
+   # TODO: For the moment we are only producing an English man page.  It would not be huge effort to produce them for
+   # non-English locales, so we should do that at some point.
+   #
+   pandocResult = run_command(
+      'pandoc',
+      join_paths(meson.current_build_dir(), 'manpage.1.md'),
```
eli-schwartz, line 1277: If you save the `configure_file()` returned value from above, you can pass it into command lines here.

Any particular reason you used run_command() instead of something like:
```meson
man_file = configure_file(
    input: configured_md_file,
    output: projectName + '.1',
    command: ['pandoc', ......],
)
install_man(man_file)
```
?

bt
```diff
@@ -0,0 +1,2152 @@
...
+   log.debug('Working directory now ' + pathlib.Path.cwd().as_posix())
+
+   #
+   # Meson can't do binary packaging, but it can create a source tarball for us via `meson dist`.  We use the following
+   # options:
+   #    --no-tests  = stops Meson doing a full build and test, on the assumption that we've already done this by the
+   #                  time we come to packaging
+   #    --allow-dirty  = allow uncommitted changes, which is needed more often than you think (eg for the automatic
+   #                     builds done when a commit is pushed up to github).  HOWEVER, note this option is only available
+   #                     in Meson 0.62 and later.
```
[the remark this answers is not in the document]
matty0ung: Thanks.  I had slightly misunderstood that.  Will fix the comment.

meson.build
```diff
@@ -0,0 +1,1449 @@
...
+# But meson.project_name() returns a Meson String, not a Python String, so there's a bit more work to do to get the same
+# result.
+#
+projectName = meson.project_name().to_lower()
+capitalisedProjectName = projectName.substring(0, 1).to_upper() + projectName.substring(1)
+
+message('⭐ Building', projectName, 'version', meson.project_version(), 'for', host_machine.system(),
+        'on', build_machine.system(), 'in', meson.project_source_root(), '⭐')
+compiler = meson.get_compiler('cpp')
+message('Using Meson', meson.version(), ', ', compiler.get_id(), 'compiler and', compiler.get_linker_id(), 'linker')
```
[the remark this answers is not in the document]
matty0ung: It's useful to have it in the main log for the github actions.  I'll add a comment.

meson.build
matty0ung: Good to know about `.full_path()`.  Will update comment.

meson.build
matty0ung: One for the future! :)

meson.build
matty0ung: Ah, that's much more elegant.  Thanks!

meson.build
matty0ung: Ah, OK, that's interesting.
AFAICT, the downside of using `dependency('xerces-c', 'XercesC')` is that, if the dependency is found natively by Meson (rather than calling out to CMake's `find_package()`), then we don't have a way to know what directory the library was found in, which is rather nice to have for passing out to the packaging scripts.  But very happy to be corrected about this!

meson.build
matty0ung: Ah, neat.  I didn't know about that.  Thanks!

eli-schwartz: It can also be useful for organizational purposes and to avoid unbounded growth of the root `meson.build` file. :)

For example, some people prefer to subdir into the src/ directory and append to a global list of sources to keep long lists of filenames separate from top-level logic about how to use them.

meson.build
matty0ung: Ah, OK, so I can use configure_file again, but this time with a custom command.  That hadn't occurred to me.

matty0ung: I hadn't noticed that `configure_file` returns a file object!

meson.build
matty0ung: Except that it doesn't work on Windows. :crying_cat_face:
Well, we'll do the new way on Mac and Linux at least.

eli-schwartz: What does dl do on Windows anyway? Is that just when building using a mingw-based compiler?

meson.build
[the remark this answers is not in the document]
eli-schwartz: What I mean is that the console log looks something like this:
```
The Meson build system
Version: 1.0.1
Source dir: /home/eschwartz/git/mesonbuild/wrapdb/subprojects/zlib-1.2.13
Build dir: /home/eschwartz/git/mesonbuild/wrapdb/subprojects/zlib-1.2.13/builddir
Build type: native build
Project name: zlib
Project version: 1.2.13
C compiler for the host machine: ccache cc (gcc 11.1.0 "cc (GCC) 11.1.0")
C linker for the host machine: cc ld.bfd 2.36.1
Host machine cpu family: x86_64
Host machine cpu: x86_64
```

It's not hidden in the debug log.

meson.build
```diff
@@ -1272,22 +1309,13 @@ if host_machine.system() == 'linux'
...
-      '--verbose',
-      '-t', 'man',
-      '-s',
-      '-o', join_paths(meson.current_build_dir(), projectName + '.1'),
+   generatedManPage = configure_file(
+      input: manPageMarkdown,
+      output: projectName + '.1',
       capture : true,
-      check : true
+      command: ['pandoc', manPageMarkdown, '--verbose', '-t', 'man', '-s'],
```
eli-schwartz, line 1326:
```suggestion
      command: ['pandoc', '@INPUT@', '--verbose', '-t', 'man', '-s'],
```

Instead of repeating manPageMarkdown, you can have Meson format the input location into the command.

meson.build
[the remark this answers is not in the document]
matty0ung: Ah, yes, I see what you mean.  You're right - no need to log it again manually.

meson.build
matty0ung: Hmm, good question.  Maybe I'll try not having it as a dependency at all on Windows.

PR #755, opened 2023-05-18
More fixes for old DBs where hop and yeast inventory_id is double instead of int

Scimmia22:


matty0ung (approved):

jimdbr:
I still see an error on inventoryId.

10:06:13.082] (1dvhehrosg) ERROR : void ObjectStore::impl::wrapAndUnmapAsNeeded(const ObjectStore::TableDefinition&, const ObjectStore::TableField&, QVariant&) Unexpected type # 6 = double in QVariant for property inventoryId , field type FieldType # 1 : ( "ObjectStore::FieldType::Int" ) , value QVariant(double, 38) , table misc , column inventory_id  [database/ObjectStore.cpp:699]
[10:06:13.093] (1dvhehrosg) ERROR : void ObjectStore::impl::wrapAndUnmapAsNeeded(const ObjectStore::TableDefinition&, const Obj

Brewtarget starts and runs fine with the default database.

Scimmia22:
I only had it on hop and yeast, looks like you have it on misc as well.

Really, databases need fixed instead of these workarounds.

jimdbr:
My database has been upgraded and merged many times over the past few years. Maybe it's time to retire the old database and start fresh. My important recipes are available in other formats and I don't mind adding them to a fresh database when needed. I see a lot of junk in my database with DB Browser. I might play with the data a bit to see what, if anything, I can recover, but it's not critical.

If I'm the only one with these errors we should close the issue and let the developers concentrate on new features. Sorry my c++ coding skills aren't good enough to assist at this time.

matty0ung:
> Really, databases need fixed instead of these workarounds.

Yes, ideally we would do this.  If you want to upload some of the problematic databases, then I can play around with some SQL to see if fixing is possible.  (We want to make sure that we fix what is broken without breaking what isn't, if you see what I mean.)  We can either do a database schema version upgrade just for the type fixes, or we can include it in the next schema update (probably when we add a lot of fields for BeerJSON).

jimdbr:
I'm OK with waiting for the next schema update. I am looking through my database to see if I can fix anything myself. I have also located and exterminated some older databases that might have contributed to my problems. The oldest I found was version 6.

jimdbr:
I also found a couple of old bits of brewtarget under /usr instead of /usr/local, so I need to do some housekeeping. I believe all the new stuff is now under /usr/local.

PR #776, opened 2023-10-19
Improve keyboard usage and reduce boilsize by deadspace.

MedicMomcilo:
When using keyboard to move slider for mash designer values are not updated.
I've switched the event that we connect to, so this usecase would be supported.

Boil size should also be reduced by the lauter deadspace in calculation.
I've simply substracted it if it's defined in the equipment.

Please take special consideration when reviewing this PR as I'm not a developer by profession.
I'm not aware if this has any impact or implications on other parts of Brewtarget.

Cheers!


matty0ung:
Thanks for this.  I'll have a proper look soon (off to a wedding this weekend) but looks good from a first glance.

matty0ung (approved):

PR #833, opened 2024-09-25
Tweaked FindBoost command in CMakeLists.txt

bebenlebricolo:
This commit repairs the configuration on Linux machines where we have a recent CMake toolset (VERSION >= 3.29) where FindBoost.cmake module is removed from standard installation.
Setting the policy CMP0167 helps ensure compatibility with older versions of CMake as well.

Note : in this commit we also uniformise Boost retrievala across platforms, not as COMPONENT but as a general dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,11 @@ cmake_policy(VERSION 3.16)
 # numbers via it (rather than manually setting individual major/minor/patch vars).
 cmake_policy(SET CMP0048 NEW)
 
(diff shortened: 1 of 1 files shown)


bebenlebricolo:
@matty0ung , In order to test on my Arch Linux machines I had to make those changes in CMakeLists.txt.
It's to ensure we can retrieve boost libraries correctly ; plus these changes uniformize the way we are retrieving boost libs on all platforms.
Feel free to review the changes as you see fit :smile:

matty0ung (approved):
It compiles, so looks like it works!

PR #871, opened 2024-11-05
Danish translation

orlamoeller:
This is an uptodate translation file for danish and I have added a line i CMakeLists.txt to include it when building executables.
There are a few words/phrases in the userinterface still in english. I assume these are not integrated in in the qt UI.
Also there is some accelerators not working. This also goes for english UI.


matty0ung:
Brilliant - many thanks! Tak!

PR #887, opened 2024-11-18
Update meson.build with danish languageentry

orlamoeller:
I added a danish language file, but I was not aware that building is done with meson. So the danish language did not make it to the new version 4.0.10. I have added the needed line to meson.build.


matty0ung:
Ah, sorry, yes, I should have spotted that.  We'll have 4.0.11 out, complete with Danish!, relatively soon.  Thanks for the PR!

matty0ung (approved):

PR #912, opened 2024-12-31
Updated danish translation

orlamoeller:
With translation of some until now untranslated phrases and a few corrections.


matty0ung:
Brilliant - many thanks!  I'll include this in the 4.0.14 release.

PR #935, opened 2025-02-26
Update recipeExtrasWidget.ui

orlamoeller:
Corrected error in date-format from dd MMM yyyy to dd MM yyyy.
The error resulted in month-number shown as 00 on the Extra tab in the recipe window. A little detail, and ideally changes in Options for date format should also be shown here. But that's a very little issue.

### DIFF
--- a/ui/recipeExtrasWidget.ui
+++ b/ui/recipeExtrasWidget.ui
@@ -171,7 +171,7 @@
          <item row="5" column="1">
           <widget class="BtOptionalDateEdit" name="dateEdit_date">
            <property name="displayFormat">
-            <string>dd MMM yyyy</string>
+            <string>dd MM yyyy</string>
            </property>
            <property name="calendarPopup">
             <bool>true</bool>


matty0ung:
Hi,

Thanks for the PR.  The current date format is deliberately the way it is, so I'm curious about what you're seeing on the Extras tab.

On my system, setting date format to `dd MMM yyyy` means you see dates with an abbreviated month name, eg, "03 Feb 2025".  The advantage of this is it's unambiguous.  If I set the date format to `dd MM yyyy`, some dates become ambiguous - eg "03 02 2025" might mean either "3 February" or "2 March" depending on where you live.

If there is a reason that `dd MMM yyyy` doesn't work in all locales, and we need to have a purely numeric date format, I would be tempted to use `yyyy-MM-dd` (aka the one true date format) to give, eg, 2025-02-03.  That said, I appreciate that's not to everyone's taste!

Probably the right answer is, exactly as you say, to have the Extras tab use whatever date format is specified in the "Options > Units" window.  But that's a bit more than a one-line change...

Anyway, let me know what you think.

orlamoeller:
My fault. I was not aware of QT's advanced date-formatting facilities. The real problem is the danish translation file , where I assumed the same failure (which is actually not a failure). So I translated MMM to MM in the translations for recipeExtrasWidget, and this is the cause of the problem, which I found out was only seen in the danish user interface.
So this PR should be withdrawn. And I will make a PR for the corrected danish translation file (which also contains a few other corrections according to new sentences in the ui and some other details, i have discovered, when using Brewtarget.

orlamoeller:
See comment above.

PR #937, opened 2025-03-01
New danish translation file with correction for causing wrong date on recipeExtrasWidget

orlamoeller:
And a few other improvements. This has no hurry. Can just be added when new version has to be made for other reasons.


matty0ung:
Excellent - many thanks!

matty0ung (approved):

Issue #940, opened 2025-03-21
Build error - TreeNodeTraits.h


jimdbr:
Fresh git pull, meson compile --clean and meson compile. Results:

432/669] Compiling C++ object libcommon_code.a.p/src_Application.cpp.o
FAILED: libcommon_code.a.p/src_Application.cpp.o 
c++ -Ilibcommon_code.a.p -I. -I.. -I../src -I../third-party/valijson/include -I../third-party/libbacktrace -I/usr/include -I/usr/local/include -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Wpedantic -std=c++20 -DBOOST_STACKTRACE_LINK -DBOOST_STACKTRACE_USE_BACKTRACE -g3 -O2 -z noexecstack -fconcepts -pie -fpie -rdynamic -fPIC -DBOOST_ALL_NO_LIB -isystem/usr/include/x86_64-linux-gnu/qt6/QtXml -DQT_XML_LIB -isystem/usr/include/x86_64-linux-gnu/qt6/QtSvg -DQT_SVG_LIB -isystem/usr/include/x86_64-linux-gnu/qt6/QtSql -DQT_SQL_LIB -isystem/usr/include/x86_64-linux-gnu/qt6/QtPrintSupport -DQT_PRINTSUPPORT_LIB -isystem/usr/include/x86_64-linux-gnu/qt6/QtWidgets -DQT_WIDGETS_LIB -isystem/usr/include/x86_64-linux-gnu/qt6/QtMultimedia -DQT_MULTIMEDIA_LIB -isystem/usr/include/x86_64-linux-gnu/qt6/QtNetwork -DQT_NETWORK_LIB -isystem/usr/include/x86_64-linux-gnu/qt6/QtGui -DQT_GUI_LIB -isystem/usr/include/x86_64-linux-gnu/qt6/QtCore -isystem/usr/include/x86_64-linux-gnu/qt6 -DQT_CORE_LIB -isystem/usr/lib/x86_64-linux-gnu/qt6/mkspecs/linux-g++ -MD -MQ libcommon_code.a.p/src_Application.cpp.o -MF libcommon_code.a.p/src_Application.cpp.o.d -o libcommon_code.a.p/src_Application.cpp.o -c ../src/Application.cpp
In file included from ../src/trees/TreeNodeBase.h:21,
                 from ../src/trees/TreeModel.h:29,
[...]

PR #941, opened 2025-03-21
Add missing include

matty0ung:
First stab at fixing https://github.com/Brewtarget/brewtarget/issues/940.


matty0ung:
This wasn't required to fix that issue in the end, but it's still a valid improvement.  This PR seems to have got stuck, so I'm going to kill it and recreate it.

PR #981, opened 2025-06-07
Fix Windows build unintentionally opening a console window

### LINKED ISSUE #976: Windows releases always open a console window in addition to the main window
Windows releases of Brewtarget always open a terminal window in addition to the main application window when launched, which is probably unintended behavior.

Image

**Tested versions:**
* 4.0.17
* 4.1.0
* 4.1.1

This is probably due to not setting the [WIN32_EXECUTABLE](https://cmake.org/cmake/help/latest/prop_tgt/WIN32_EXECUTABLE.html#prop_tgt:WIN32_EXECUTABLE) property in `CMakeLists.txt`.  Possibly that could be added in this section:

https://github.com/Brewtarget/brewtarget/blob/0c32495bc27c91b77ae57aa17c17d0698a29109b/CMakeLists.txt#L534-L614

I'm still getting set up to test building this locally, but I'm thinking maybe this would help address it:

```
# This section for context
   get_target_property(_qmake_executable Qt6::qmake IMPORTED_LOCATION)
   get_filename_component(QT_BIN_DIR "${_qmake_executable}" DIRECTORY)
   message("QT_BIN_DIR = ${QT_BIN_DIR}")

# Proposed solution
  if(CMAKE_BUILD_TYPE STREQUAL "Release")
    set_property(TARGET main PROPERTY WIN32_EXECUTABLE true)
  endif()
```

taylor-madeak:
From Meson docs:

https://mesonbuild.com/Reference-manual_functions.html#executable_win_subsystem

The default value for `win_subsystem` is `'console'`, which resulted in the unintended behavior of a console window being opened in addition to the main Brewtarget GUI.  Setting `win_subsystem: windows` resolves #976.

### DIFF
--- a/CHANGES.markdown
+++ b/CHANGES.markdown
@@ -13,9 +13,6 @@ happens, so I'm now setting it to a slightly arbitrary time early in the morning
 * TODO: Additional methods for calculating IBU
 * We'll list other new features here...
 
-### Bug Fixes
-* Windows version opens a console in addition to the main window [976](https://github.com/Brewtarget/brewtarget/issues/976)
-
 ## v4.1.2
 Bug fixes and minor enhancements.
 
@@ -24,6 +21,7 @@ Bug fixes and minor enhancements.
 
 ### Bug Fixes
 * Crash when deleting a water profile [974](https://github.com/Brewtarget/brewtarget/issues/974)
+* Windows version opens a console in addition to the main window [976](https://github.com/Brewtarget/brewtarget/issues/976)
 
 ### Release Timestamp
 Mon, 2 Jun 2025 04:01:02 +0100
--- a/CHANGES.markdown
+++ b/CHANGES.markdown
@@ -13,6 +13,9 @@ happens, so I'm now setting it to a slightly arbitrary time early in the morning
 * TODO: Additional methods for calculating IBU
 * We'll list other new features here...
 
+### Bug Fixes
+* Windows version opens a console in addition to the main window [976](https://github.com/Brewtarget/brewtarget/issues/976)
+
 ## v4.1.2
 Bug fixes and minor enhancements.
 
--- a/meson.build
+++ b/meson.build
@@ -1720,7 +1720,7 @@ mainExecutable = executable(mainExecutableTargetName,
                             link_with : commonCodeStaticLib,
                             kwargs : platformSpecificArgs,
                             install : true,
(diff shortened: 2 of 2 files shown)


matty0ung:
Did you have any luck with the equivalent fix in CMake?  If so, let's include that into this PR, and then I'll merge them into the main codebase together.  If not, we'll tag it as a "known issue", and I'll just merge the Meson fix.

taylor-madeak:
I added my proposed fix to `CMakeLists.txt`, but in testing I found that CMake doesn't seem to work for building this in my MSYS2 environment.  The issue is that the `cmake ..` command completes without error, but doesn't generate a Makefile in my build directory.  It generates a whole bunch of other stuff, but no Makefile.

I can work on it some more today, but advice on how to troubleshoot the issue would be appreciated. :)

matty0ung:
On Windows, there are a couple of things to do:

- Before you run CMake for the first time, you want to run `./configure` in the top-level directory.  This creates the `build` directory and sets a few bits of CMake config up.
- Then, once you're inside the `build` directory, you want to run `cmake --build .`.  If that still fails mysteriously, try `cmake --build . --verbose`, which give a bit more diagnostic about what's going on.

taylor-madeak:
Thank you for that guidance, I will test that in the next couple of hours and update the PR accordingly.  I was following the instructions on the [Development: Getting Started](https://github.com/Brewtarget/brewtarget/wiki/Development:-Getting-Started) page, which look like they need to be updated based on your guidance just now. 😀

matty0ung:
> Thank you for that guidance, I will test that in the next couple of hours and update the PR accordingly. I was following the instructions on the [Development: Getting Started](https://github.com/Brewtarget/brewtarget/wiki/Development:-Getting-Started) page, which look like they need to be updated based on your guidance just now. 😀

Ah, yes, fair point about the wiki page!  I've made a small edit to correct this point.

taylor-madeak:
I was able to get a successful build with CMake, thank you.  No changes are required for CMake, it looks like you already addressed this specific issue in #655.  I think this can now be classified as fixing a regression related to switching build tools?

taylor-madeak:
I was able to get a successful build with CMake, thank you.  No changes are required for CMake, it looks like you already addressed this specific issue in #655.  I think this can now be classified as fixing a regression related to switching build tools?

PR #1048, opened 2025-11-29
Updated danish translation

orlamoeller:
Here is an updated danish translation file to include in next build.


matty0ung:
Many thanks!

Issue #1085, opened 2026-03-15
Deleting recipe locks associated linked processes from deletion


SideSlide:
If you delete a recipe you cannot delete the associated boil, fermentation, mash profile, and other(?) items which it used.

matty0ung:
Thanks for flagging this up.  It's a missing bit of filtering.  I'll include a fix in the next release.

SideSlide:
Resolved

matty0ung:
Fix included in [Brewtarget 5.0.4](https://github.com/Brewtarget/brewtarget/releases/tag/v5.0.4).

PR #1087, opened 2026-03-16
Allow deletion of an item when only deleted recipes use it

matty0ung:
Should address https://github.com/Brewtarget/brewtarget/issues/1085.


SideSlide:
I built this PR and the problem looks to be resolved

Issue #1086, opened 2026-03-16
Stock Purchase button showing on incorrect catalog pages


SideSlide:
The stock purchase button is showing in the upper left corner of the catalog / database UI for equipment, mashes, boils, fermentations, water, and styles.

matty0ung:
Thanks for flagging this up.  I'll take a look.  The button obviously shouldn't show at all on those windows.

SideSlide:
Fixed

matty0ung:
Fix included in [Brewtarget 5.0.4](https://github.com/Brewtarget/brewtarget/releases/tag/v5.0.4).

PR #1091, opened 2026-03-16
Fix erroneous stock purchase button on non-ingredient catalog windows

matty0ung:
Addresses https://github.com/Brewtarget/brewtarget/issues/1086


SideSlide:
Just built this, and looks like it's all cleared up.

PR #1111, opened 2026-05-03
Make the build script work on Debian sid

trahay:
Hi,
When building brewtarget on my Debian sid laptop, I had to change a few things to make `bt` work:
- add missing package dependencies
- fix `getLinuxDistroInfo` because `lsb_release -a` returns 'n/a' on Debian sid, which makes parsing the version number crash.


matty0ung (changes requested):
Thanks for this!  Always good to get things running on more distros.

(I had run `lsb_release -a` on my Debian Trixie box, where it works fine, but didn't know it's problematic on Sid.)

I think you want to lose the 'break' line as it, err, breaks things. :smile_cat:  Otherwise looks good.

